PR #1, opened 2021-09-13
initial version of configure_kernel_ima_module

kkaarreell:
This test will be used as a setup step to enable kernel IMA on a tested system.


install/configure_kernel_ima_module/test.sh
```diff
@@ -0,0 +1,39 @@
...
+
+COOKIE=/var/tmp/configure-kernel-ima-module-rebooted
+TESTFILE=/var/tmp/configure-kernel-ima-module-test$$
+
+rlJournalStart
+
+  if [ ! -e $COOKIE ]; then
+    rlPhaseStartSetup "pre-reboot phase"
+        rlRun "grubby --info DEFAULT | grep '^args'"
+        rlRun "grubby --update-kernel DEFAULT --args 'ima=$IMA_STATE ima_policy=$IMA_POLICY'"
```
sopos, line 16: Shouln't we boot it with `ima_appraise=fix` first?

kkaarreell: I think this is used for local attestation and we do not need it for keylime. Anyway, we can add it later if needed.

install/configure_kernel_ima_module/test.sh
```diff
@@ -0,0 +1,39 @@
...
+  else
+    rlPhaseStartTest "post-reboot IMA test"
+        rlRun -s "cat /proc/cmdline"
+        rlAssertGrep "ima=$IMA_STATE" $rlRun_LOG
+        rlAssertGrep "ima_policy=$IMA_POLICY" $rlRun_LOG
+        rlRun "rm $COOKIE"
+
+        if [ "$IMA_STATE" == "on" -o "$IMA_STATE" == "1" ]; then
+          rlRun "touch $TESTFILE && cat $TESTFILE && rm $TESTFILE"
+	  rlRun "grep $TESTFILE /sys/kernel/security/ima/ascii_runtime_measurements"
```
sergio-correia, line 34: You may want to adjust the indent here, but otherwise it looks good.

sergio-correia (approved):

install/configure_kernel_ima_module/test.sh
sopos: ima_appraise=enforce is the default value so I'm not sure what is the real impact of it.

PR #11, opened 2021-11-04
tests Multihost/basic-attestation

kkaarreell:
Adding first functional version of multihost test doing basic attestation, using the following setup:
SERVER1 - Verifier
SERVER2 - Registrar
SERVER3 - Agent + Tenant
Custom certificates are generated for the test and certificate verification is enabled.


kkaarreell:
Hello @sergio-correia @mpeters @sopos , would you have any feedback on this MR?

sopos (approved):

sergio-correia (approved):
Great work, it looks good.

sergio-correia:
Only feedback I have here is that it would be interesting to use `keylime_ca` to generate certs, as a way to test that tool as well. We can do it in the future, probably in a different test.

kkaarreell:
> Only feedback I have here is that it would be interesting to use `keylime_ca` to generate certs, as a way to test that tool as well. We can do it in the future, probably in a different test.

To my understanding these certs are generated in a default configuration which is used by the singlehost test.
If you mean generating certificates for the multihost scenario, true, that would be also possible and better doing that in another test.

Thank you for the review.

PR #12, opened 2021-11-11
add swtpm support on Fedora

kkaarreell:


kkaarreell:
I have tested it on F34 and it works well.
On F35 there is the following failure
```
# keylime_tenant -c cvlist
Using config file /etc/keylime.conf
2021-11-11 11:15:41.183 - keylime.tpm - INFO - TPM2-TOOLS Version: 5.2
2021-11-11 11:15:41.185 - keylime.tenant - INFO - Setting up client TLS in /var/lib/keylime/cv_ca
2021-11-11 11:15:41.186 - keylime.tenant - WARNING - Using default UUID d432fbb3-d2f1-4a97-9ef7-75bd81c00000
/usr/lib/python3.10/site-packages/urllib3/connectionpool.py:1013: InsecureRequestWarning: Unverified HTTPS request is being made to host '127.0.0.1'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
2021-11-11 11:15:41.210 - keylime.tenant - INFO - Status command response: 500. Unexpected response from Cloud Verifier 127.0.0.1 on port 8881. <Response [500]>
```
but there is nothing suspicious in the verifier.log. This should not be a test code issue though.

functional/basic-attestation-on-localhost/test.sh
```diff
@@ -42,7 +42,7 @@ rlJournalStart
         rlRun "keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u $AGENT_ID -f excludelist.txt --allowlist allowlist.txt --exclude excludelist.txt -c add"
         sleep 5
         rlRun -s "keylime_tenant -c cvlist"
-        rlAssertGrep "{'agent_id': '$AGENT_ID'}" $rlRun_LOG
+        rlAssertGrep "{'code': 200, 'status': 'Success', 'results': {'uuids':.*'d432fbb3-d2f1-4a97-9ef7-75bd81c00000'" $rlRun_LOG -E
```
mpeters, line 45: Why was this changed to be hard coded vs using the `$AGENT_ID` ?

setup/configure_tpm_emulator/test.sh
```diff
@@ -48,16 +53,64 @@ User=tss
...
+[Service]
+Type=dbus
+BusName=com.intel.tss2.Tabrmd
+ExecStart=/usr/sbin/tpm2-abrmd --tcti=swtpm
+User=tss
+
+[Install]
+WantedBy=multi-user.target
+_EOF"
+            # now we need to build custom selinux module making swtpm_t a permissive domain
```
mpeters, line 94: Is this something we should report upstream?

functional/basic-attestation-on-localhost/test.sh
kkaarreell: Mistake. Fixed now.

setup/configure_tpm_emulator/test.sh
kkaarreell: Maybe. Upstream swtpm is not using systemd unit file, probably on purpose due to many various ways how swtpm can be configured and started. I assume this might the reason why SELinux module doesn't have rules for transition from `init_t` to `swtpm_t` domain. This is one thing added by my module. However, even after adding these rules I have noticed some AVCs. Partially due to my configuration (accessing files in /var/lib/tmp), partially unrelated. I will try to gather more logs and report an issue upstream to find out.

PR #15, opened 2021-11-16
packit,testing-farm: simplify and use skip_build

TomasTomecek:
no need to have spec file and the boilerplate packit.yaml any more


kkaarreell:
Thank you @TomasTomecek .

PR #16, opened 2021-11-23
enable functional tests in plan keylime-tests-github-ci.fmf

kkaarreell:


kkaarreell:
Test failed. Despite https://github.com/psss/tmt/pull/950 has been merged we need to wait for new tmt release. That should happen by the end of November.

kkaarreell:
We can ignore test failures since the only important thing to verify here is that new tmt preserve test order.. and it does.
Failures are caused by different issues.

thrix:
Yeah, the poor order is removed in the HTML conversion seems, the order looks correct in xunit:
http://artifacts.dev.testing-farm.io/f083abdf-bf9a-49f5-bdd3-2b08ac1886a5/results-junit.xml

PR #20, opened 2022-01-07
[WIP] Enable code coverage measuring for keylime services

kkaarreell:
Starts keylime services via "coverage run" when COVERAGE environment variable is defined and non-empty.
This enable gathering code coverage metrics.

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -30,9 +30,6 @@
 # Include Beaker environment
 . /usr/share/beakerlib/beakerlib.sh || exit 1
 
-# set monitor mode so we can kill the background process
-# https://unix.stackexchange.com/questions/372541/why-doesnt-sigint-work-on-a-background-process-in-a-script
-[ -n "$COVERAGE" ] && set -m
 
 true <<'=cut'
 =pod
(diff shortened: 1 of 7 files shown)


kkaarreell:
Requires coverage script to be available by doing "pip3 install coverage". I think we can implement another setup task to do the required setup.
Eventually, I am considering to replace COVERAGE env variable with a dedicated file storing coverage script parameters (stored by the setup task) so that usage of coverage script is auto detected. This will be probably subject of future update, for now I would just want to get some feedback.

kkaarreell:
Currently we run keylime_tenant via expect (since entering a password/key is necessary) and therefore running it via coverage script would be a bit more tricky.

kkaarreell:
> Currently we run keylime_tenant via expect (since entering a password/key is necessary) and therefore running it via coverage script would be a bit more tricky.

This is addressed in the new commit by introducing `lime_keylime_tenant` wrapper.

sopos (approved):
It seems to me reasonable. I would like to see also the report interpretation, tough.

LGTM

kkaarreell:
> It seems to me reasonable. I would like to see also the report interpretation, tough.

@sopos I have added `setup/generate_coverage_report` task.

sopos:
Can we also update plans/keylime-tests-github-ci.fmf to see the results in the CI?

kkaarreell:
> Can we also update plans/keylime-tests-github-ci.fmf to see the results in the CI?

I am not sure it would be helpful. We would see only overall coverage per file but I am not able to attached the HTML report (or an archive of it).

kkaarreell:
> > Can we also update plans/keylime-tests-github-ci.fmf to see the results in the CI?
> 
> I am not sure it would be helpful. We would see only overall coverage per file but I am not able to attached the HTML report (or an archive of it).

I will enable it for this tests repo so we can see that it is working with future test updates.
https://github.com/RedHat-SP-Security/keylime-tests/pull/22

PR #30, opened 2022-01-19
Improve logging when service stop failed

kkaarreell:
Attachments are not available in Packit CI and therefore it is handy to see tail of a log when service cannot be stopped.


kkaarreell:
I *HOPE* this is the last update in this piece of code, at least for the next several weeks. :-)

PR #32, opened 2022-01-20
Replace ibmswtpm2 with swtpm on RHEL

kkaarreell:
RHEL-9 already contains swtpm so we do not need to use custom build of ibmswtpm2.


sergio-correia (approved):
LGTM, test still passing here.

PR #34, opened 2022-01-20
Added instruction for multi-host test execution

kkaarreell:


kkaarreell:
Could you please test it to confirm it works fine?

sergio-correia:
> Could you please test it to confirm it works fine?

Sure thing, I am on it and will report back here.

TESTING.md
```diff
@@ -69,6 +69,38 @@ tests are finished. The command below will give you a shell after all tests are
...
+For `Multihost/basic-attestation` test we need at least 3 test systems, i.e. 3 terminals.
+
+On terminal 1:
+```
+$ git clone https://github.com/RedHat-SP-Security/keylime-tests.git
+```
+
+On all terminals:
+```
+$ cd keylime-tests
```
sergio-correia, line 85: This is confusing. From the previous instruction, we only cloned the repo in terminal 01. Should we do the same for all the terminals, or should we simply create a `keylime-tests` directory in the other ones?

kkaarreell: All 3 terminals are on a local system so it is enough to clone the repo once and enter it in each terminal. I will rephrase it.

kkaarreell:
> Could you please test it to confirm it works fine?

Now I know it won't work since with `provison virtual` all 3 systems will have the same IP and would be most likely on a separate network. I will try to find out how this can be changed.

TESTING.md
sergio-correia: Ahaa, now it makes sense. Inded, I misunderstood the beginning of the instructions. Here I had 3 different systems, then 3 terminals, one for each system :)

kkaarreell: Updated.

kkaarreell:
> > Could you please test it to confirm it works fine?
> 
> Now I know it won't work since with `provison virtual` all 3 systems will have the same IP and would be most likely on a separate network. I will try to find out how this can be changed.

Should be fixed now with updated tmt command.

kkaarreell:
Unfortunately, it still won't work since we do not have sync mechanism available outside RH.
Internally, we could either install the required dependencies or provision it using 1minutetip. 
But I am afraid that for now we do not have an alternative (unless we would be installing restraint).

TESTING.md
```diff
@@ -63,12 +63,47 @@ tests are finished. The command below will give you a shell after all tests are
...
+$ tmt run -vvv prepare discover -h fmf -t 'configure_tpm_emulator' -t 'install_upstream_keylime' -t 'Multihost/basic-attestation' provision -h virtual -i Fedora-35 execute login finish
```
sergio-correia, line 89: Perhaps it might be useful to mention that during this step it will download an image for using in the test? And actually, that it might be wise to wait for the image to be downloaded before going to the 2nd and 3rd terminals to provision the other systems, as the image can be reused once it's downloaded.

In my test, it was a Fedora-Cloud-Base-35-1.2.x86_64.qcow2 file, and I don't know how it works, but apparently it was getting it from a really slow mirror, so it was taking very long, this download. I actually ended up stopping it and then downloading it myself from a different mirror, once I found it was expected to be at `/var/tmp/tmt/testcloud/images`.

Another point that might be useful to mention is that we need libvirtd running, or elsewe will get errors like 
```
progress: booting...
libvirt: XML-RPC error : Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory
libguestfs: error: could not connect to libvirt (URI = qemu:///system): Failed to connect socket to '/var/run/libvirt/virtqemud-sock': No such file or directory [code=38 int1=2]
```

Another issue I had the first time (this was a "brand new" minimal Fedora 35 system) was with the tmt provision stuff. I needed some specific package to be able to use tmt provision virtual; package probably is named tmt-virtual, if I recall correctly. I found out there was a tmt-all and installed this one instead, though.

TESTING.md
```diff
@@ -63,12 +63,47 @@ tests are finished. The command below will give you a shell after all tests are
...
+```
+
+Multihost test won't be run properly but at least tmt will install all test requires and do the setup.
```
sergio-correia, line 92: Do you have a typo here, perhaps, and mean that tmt will install the test requirements?

TESTING.md
```diff
@@ -63,12 +63,47 @@ tests are finished. The command below will give you a shell after all tests are
...
+
+Once we have an interactive shell on every test system we can proceed with test execution.
+
+On every test system:
+```
+# cd /var/tmp/tmt/run-*/plans/keylime-tests-github-ci/discover/default/tests/Multihost/basic-attestation/
+# ## find out hostname or IP of each test system, e.g. with "hostname -I"
```
sergio-correia, line 99: Hmm.  I wonder if I did something wrong here -- all 3 interactive shells show the same hostname and IP(v4), apparently:
`interactive shell #1`
```
[root@testcloud tree]# hostname -I
172.17.2.15 10.0.2.15 fec0::5054:ff:fe54:b399 fec0::c1b2:38:2b6e:4682
```

`interactive shell #2`
```
[root@testcloud tree]# hostname -I
172.17.2.15 10.0.2.15 fec0::5054:ff:fe80:4302 fec0::21b2:1242:25a1:f8cd
```

`interactive shell #3`
```
[root@testcloud tree]# hostname -I
172.17.2.15 10.0.2.15 fec0::5054:ff:fe03:1dc9 fec0::175:71:5c26:eae1
```

sergio-correia: And it seems I was using the old tmt command line... let me retest with the extra `-c system`

sergio-correia: And apparently it does; at least I see already a different address for each system:
```
progress: booting...
ip: 192.168.124.163
```
```
progress: booting...
ip: 192.168.124.116
```
```
progress: booting...
ip: 192.168.124.162
```
OK, I guess I reached the point of your last comment, that it still won't work because we don't have a sync mechanism.

TESTING.md
kkaarreell: Hi @sergio-correia , thank you for a detailed review. I have updated doc with additional instructions. Also, please note that the top level section 'Running tests in a virtual system using tmt' contains more details regarding the required workstation setup.

TESTING.md
kkaarreell: Fixed.

TESTING.md
kkaarreell: Thank you. I am working on a custom sync mechanism so it can be used by folks outside RH.

TESTING.md
sergio-correia: Yeah, now I see that I missed that section. Thanks!

TESTING.md
```diff
@@ -57,18 +60,58 @@ Then you can run a test plan e.g. on F35 system.
...
+```
+
+In the next step we will start 3 virtual systems using `tmt`.
+
+In case you do not have Fedora-35 image already downloaded (in `/var/tmp/tmt/testcloud/images`) you may want to run commands below in the 1st terminal first to avoid simultaneous downloads.
+
+On all terminals:
+```
+$ cd keylime-tests
+$ tmt run -vvv prepare discover -h fmf -t 'configure_tpm_emulator' -t 'install_upstream_keylime' -t 'Multihost/basic-attestation' provision -h virtual -i Fedora-35 execute login finish
```
sergio-correia, line 96: `-c system` should be here as well, no?

TESTING.md
```diff
@@ -57,18 +60,58 @@ Then you can run a test plan e.g. on F35 system.
 # tmt run -vvv prepare discover provision -h virtual -i Fedora-35 execute finish
 ```
 
-The above command will dispose the system when tests are finished.
+The above command will download Fedora-35 image and use it for a newly created virtual
+system. Also, due to `finish` command the system will disposde when tests are executed.
```
sergio-correia, line 64: typo: dispose

TESTING.md
kkaarreell: For singlehost tests it is not necessary. But maybe it would be better to keep the cmdline consistent. I will add it.

sergio-correia: But here we are talking about the multi-host ones, no? I see that we have `-c` in the previous section, which should be the one on singlehost tests. But I agree that we should keep it consistent, if it will not break things :)

kkaarreell: Aaah, you are right! Fixed now.

TESTING.md
kkaarreell: Fixed

kkaarreell:
> Unfortunately, it still won't work since we do not have sync mechanism available outside RH. 

This is now addressed in https://github.com/RedHat-SP-Security/keylime-tests/pull/35

PR #35, opened 2022-01-21
Add Library/sync and adjust multihost test for it

kkaarreell:
Implements Library/sync so that multihost tests can be run outside of Beaker/Restraint environment.


kkaarreell:
@sopos @sergio-correia Could you please verify it works for your when following updated instructions from
https://github.com/RedHat-SP-Security/keylime-tests/blob/docmulti/TESTING.md

Thank you.

sergio-correia:
Using this branch and the updated instructions from the other PR makes all 3 systems complete the tests with a PASS, after the `XTRA=1 ./test.sh`

PR #47, opened 2022-02-02
Enable python tracing for keylime services

kkaarreell:


kkaarreell:
Hi @sergio-correia , do you think this change is worth doing? Would that be beneficial? Adding trace log to a service log may possibly conflict with some tests that are checking tail of a service log. OTOH, I expect this to be used only by advanced users.

sergio-correia:
> Hi @sergio-correia , do you think this change is worth doing? Would that be beneficial? Adding trace log to a service log may possibly conflict with some tests that are checking tail of a service log. OTOH, I expect this to be used only by advanced users.

I think it may be useful sometimes. Is there a way to enable/disable the tracing? Or to send its data to a separate file, so that it doesn't cause the conflict you mentioned?

Unrelated, but still about logs, I am wondering if it will require a big change in the tests if we start using the systemd unit files to start the services, instead of running them directly.

kkaarreell:
> I think it may be useful sometimes. Is there a way to enable/disable the tracing? 

This is at the top of the patch. It is disabled by default and enabled either by exporting `limeTRACE` variable or creating a file `/var/tmp/limeLib/trace`.

> Or to send its data to a separate file, so that it doesn't cause the conflict you mentioned?

Python trace module does send it to stdout and there is no option to redirect it. It would be possible only by patching the module itself. But I expect this to be used rarely so maybe let's merge it just as an example how it can be used and consider more updates once we have clear need.

> Unrelated, but still about logs, I am wondering if it will require a big change in the tests if we start using the systemd unit files to start the services, instead of running them directly.

I have on purpose designed access to logfile through a command (not variable), e.g.
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/Library/test-helpers/lib.sh#L896
so that we can eventually construct the log file on demand (by storing `journalctl` output).
So I hope that the only changes necessary would be in `test-helpers` library itself, in particular in functions starting and stopping services and providing a logfile. Of course, for coverage measurement and tracing we would have to tweak the unit file but for simple service start/stop required changes should be small.

kkaarreell:
I am closing this PR. It is probably not worth applying right now, OTOH at least it suggests a way how such a functionality could be applied locally when necessary.

PR #50, opened 2022-02-04
Multi

kkaarreell:


kkaarreell:
I have just confirmed with `tmt` that the update tests passes both in 3 hosts and 4 hosts scenario.

PR #52, opened 2022-02-15
Improve test execution on CentOS Stream using tmt

kkaarreell:


plans/keylime-tests-github-ci.fmf
```diff
@@ -1,5 +1,12 @@
 summary:
   Tests used by Packit/TFT CI on Github
+adjust:
+  - when: distro == centos-stream-9
+    prepare+:
```
sergio-correia, line 5: Is this `+` sign expected here? (also in the next statement)

kkaarreell: Yes. It means that prepare steps should not be overridden, just extended.

sergio-correia: Nice, thanks for the info.

sergio-correia (approved):

PR #53, opened 2022-02-15
Fix coverage collection

sergio-correia:


kkaarreell:
@sergio-correia This is great, can we merge it? Btw, I have asked on a python forum for an advice but so far no response.
http://lists.idyll.org/pipermail/testing-in-python/2022-February/007386.html

sergio-correia:
> @sergio-correia This is great, can we merge it? Btw, I have asked on a python forum for an advice but so far no response. http://lists.idyll.org/pipermail/testing-in-python/2022-February/007386.html

Sure, we can merge it. And thanks for the link, I will watch that forum thread as well.

kkaarreell (approved):

PR #54, opened 2022-02-15
setup/install_upstream_keylime: Install python shim

ansasaki:
Add a step to install the python shim used to run python revocation actions.

This allows the rust agent to run and pass the `functional/basic-attestation-on-localhost`.


kkaarreell:
Would other tests pass too? Would it make sense already to create rust specific test plan.

ansasaki:
> Would other tests pass too? Would it make sense already to create rust specific test plan.

I'll try to run all tests and see. If I can make them to pass I'll add the change to the plan to uncomment the rust agent tests.

kkaarreell:
Hi @ansasaki , do you know what's the problem in `cargo build`?

ansasaki:
I think it is the change in provides for the `tpm2-tss-devel` package

The passing jobs used:
Fedora 34: https://koji.fedoraproject.org/koji/rpminfo?rpmID=26147126
Fedora 35: https://koji.fedoraproject.org/koji/rpminfo?rpmID=27152738

They have:
```
pkgconfig(tss2-sys) = 3.1.0
```
One of the failing tests used:
Fedora 36: https://koji.fedoraproject.org/koji/rpminfo?rpmID=29483497

Which have:
```
pkgconfig(tss2-sys)
```

functional/basic-attestation-on-localhost/test.sh
```diff
@@ -118,7 +118,19 @@ rlJournalStart
         rlRun "limeWaitForVerifier"
         rlRun "limeStartRegistrar"
         rlRun "limeWaitForRegistrar"
-        rlRun "limeStartAgent"
+        # set service environment variables
+        if [ -f /usr/lib/systemd/system/keylime_agent.service -o -f /etc/systemd/system/keylime_agent.service ]; then
```
kkaarreell, line 122: I am thinking whether it won't be better to move this kind of setup to test-helpers library itself, somewhere here
https://github.com/RedHat-SP-Security/keylime-tests/blob/3de7e180836511a2a1f0ecd4cd95032611225726/Library/test-helpers/lib.sh#L1299

ansasaki: Yes, it would be better if this initialization was centralized and we could start the agent with a simple `limeStartAgent` like other services.

kkaarreell:
Hi @ansasaki , I have merged all the abrmd related changes. I think that all test updates from this PR can be removed and we can preserve just updates for `/setup/install_upstream_rust_keylime/test.sh` and `/plans/rust-keylime-tests.fmf`.
Are all tests from the new plan supposed to pass? I tried running them but `/functional/basic-attestation-on-localhost` has failed.

ansasaki:
> Hi @ansasaki , I have merged all the abrmd related changes. I think that all test updates from this PR can be removed and we can preserve just updates for `/setup/install_upstream_rust_keylime/test.sh` and `/plans/rust-keylime-tests.fmf`. Are all tests from the new plan supposed to pass? I tried running them but `/functional/basic-attestation-on-localhost` has failed.

Yes, all tests included in the plan were supposed to pass.

functional/basic-attestation-without-mtls/test.sh
```diff
@@ -55,23 +46,15 @@ _EOF"
...
 
     rlPhaseStartTest "Check that empty script_payload allows the agent to start"
         rlRun "limeUpdateConf cloud_agent payload_script ''"
-        if [ -f /usr/lib/systemd/system/keylime_agent.service -o -f /etc/systemd/system/keylime_agent.service ]; then
-            rlRun "systemctl daemon-reload"
-            rlRun "limeStartAgent"
-        else
-            rlRun "RUST_LOG=keylime_agent=trace limeStartAgent"
-        fi
+        rlRun "systemctl daemon-reload"
```
kkaarreell, line 56: Shouldn't be necessary anymore.

functional/basic-attestation-without-mtls/test.sh
```diff
@@ -87,12 +70,8 @@ _EOF"
         rlRun "limeStopAgent"
         rlRun "limeUpdateConf cloud_agent enable_insecure_payload True"
         rlRun "limeUpdateConf cloud_agent payload_script 'autorun.sh'"
-        if [ -f /usr/lib/systemd/system/keylime_agent.service -o -f /etc/systemd/system/keylime_agent.service ]; then
-            rlRun "systemctl daemon-reload"
-            rlRun "limeStartAgent"
-        else
-            rlRun "RUST_LOG=keylime_agent=trace limeStartAgent"
-        fi
+        rlRun "systemctl daemon-reload"
```
kkaarreell, line 73: Shouldn't be necessary anymore.

kkaarreell:
OK, probably we need to update the test a bit. The log says:
```
:: [ 10:44:27 ] :: [   PASS   ] :: Command 'rlWaitForCmd 'tail /var/tmp/limeLib/agent.log | grep -q "Executing revocation action"' -m 10 -d 1 -t 10' (Expected 0, got 0)
:: [ 10:44:28 ] :: [  BEGIN   ] :: Running 'tail /var/tmp/limeLib/agent.log | grep 'Executing revocation action local_action_modify_payload''
May 06 10:44:27 ip-172-31-19-250.us-east-2.compute.internal keylime_agent[36012]:  INFO  keylime_agent::revocation      > Executing revocation action local_action_modify_payload
:: [ 10:44:28 ] :: [   PASS   ] :: Command 'tail /var/tmp/limeLib/agent.log | grep 'Executing revocation action local_action_modify_payload'' (Expected 0, got 0)
:: [ 10:44:28 ] :: [  BEGIN   ] :: Running 'tail /var/tmp/limeLib/agent.log | grep 'A node in the network has been compromised: 127.0.0.1''
:: [ 10:44:28 ] :: [   FAIL   ] :: Command 'tail /var/tmp/limeLib/agent.log | grep 'A node in the network has been compromised: 127.0.0.1'' (Expected 0, got 1)
:: [ 10:44:28 ] :: [   PASS   ] :: File /var/tmp/test_payload_file should not exist 
```
Here `rlWaitForCmd` succeeded at a moment when `Executing revocation action` appeared in a log, however a further message `A node in the network has been compromised` hasn't been logged yet.
I would propose to update `rlWaitForCmd` so it checks for `A node in the network has been compromised` (last one) message instead. Also, it may be worth updating also other tests doing similar check.

ansasaki:
> Here `rlWaitForCmd` succeeded at a moment when `Executing revocation action` appeared in a log, however a further message `A node in the network has been compromised` hasn't been logged yet. I would propose to update `rlWaitForCmd` so it checks for `A node in the network has been compromised` (last one) message instead. Also, it may be worth updating also other tests doing similar check.

I'll add a commit to make the changes you proposed.

kkaarreell:
I think I know now what is going on. Just need some time find the right fix. The problem is that we call `limeAgentLogfile` just once while we should run it with every wait iteration (in order to populate the file with the latest content). I will report back.

Multihost/basic-attestation/test.sh
```diff
@@ -310,9 +310,7 @@ _EOF"
         rlRun "$TESTDIR/keylime-bad-script.sh"
         rlRun "limeWaitForAgentStatus $AGENT_ID '(Failed|Invalid Quote)'"
         # give the revocation notifier a bit more time to contact the agent
-        rlRun "rlWaitForCmd 'tail $(limeAgentLogfile) | grep -q \"Executing revocation action\"' -m 20 -d 1 -t 20"
-        rlRun "tail $(limeAgentLogfile) | grep 'Executing revocation action local_action_modify_payload'"
-        rlRun "tail $(limeAgentLogfile) | grep 'A node in the network has been compromised: ${AGENT_IP}'"
+        rlRun "rlWaitForCmd 'tail $(limeAgentLogfile) | grep -q \"A node in the network has been compromised: ${AGENT_IP}\"' -m 20 -d 1 -t 20"
```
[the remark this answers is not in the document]
ansasaki: Are those necessary? I removed them because the line tested in `rlWaitForCmd` is printed by the script, so there is no need to check for the other lines to make sure the script was executed.

I'll replace `$(limeAgentLogfile)` with `\$(limeAgentLogfile)` on tail command.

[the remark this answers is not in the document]
kkaarreell: Maybe not necessary, but I assume that some log messages are important and we would like to make sure they are logged (like the message about an agent failing attestation). OTOH, we do not need to have these in all tests, one test may be enough.

kkaarreell:
@ansasaki Thank you, I am happy with the result. Can we merge it?

ansasaki:
> @ansasaki Thank you, I am happy with the result. Can we merge it?

Yes, let's merge it!

PR #61, opened 2022-02-24
Add test basic-attestation-with-unpriviledged-agent

kkaarreell:


kkaarreell:
Failures on distribution keylime are expected due to missing features&fixes.

kkaarreell:
@mpeters FYI, in this test case I am using setup with keylime_agent using separate `/var/lib/keylime-agent` directory.

PR #67, opened 2022-03-17
Add test /functional/tenant-allowlist-sanity

kkaarreell:


sergio-correia:
@kkaarreell: unrelated question: I was looking at the testing farm results and I was expecting to see the commit/revision the `/setup/install_upstream_keylime` test was using when cloning the upstream repo, but I could not find it. Is it there somehow?

kkaarreell:
@sergio-correia Commit is not being used for Testing farm / `tmt`. Instead, the repository (i.e. branch in PR) is uploaded to a test system and `tmt` command is issued from that directory. This is the reason why we are doing this
https://github.com/keylime/keylime/blob/master/packit-ci.fmf#L7
so we can get to keylime sources later in `/setup/install_upstream_keylime`.
Additionally, it is easy to run keylime-tests from local keylime branch.

kkaarreell:
I think we can add this, although sometimes the information won't be available, keylime may not be a repo but extracted archive. Also, files may be modified on a test system.. Maybe we should print something like `git status` to check for changes.

sergio-correia:
By the way, you can now revert https://github.com/RedHat-SP-Security/keylime-tests/pull/67/commits/ece3341f

PR #68, opened 2022-03-18
Run upstream bits as keylime user

kkaarreell:
Enables service file installation in the /setup/install-upstream-keylime task.
Also, since current service files start all services under `keylime` user we need to do some adjustments
in services files and `lib.sh` so that coverage measurements remain functional.


kkaarreell:
Test results looks good. Failures on distribution keylime are expected as these are caused by missing patches.

sergio-correia (approved):

kkaarreell:
Let's make one more test run with current upstream bits.

PR #69, opened 2022-03-23
Update basic-attestation-on-localhost to use custom certificates

kkaarreell:
There are already several tests that are using certificates generated by verifier but there is no test using custom certificate except the Multihost one. As a result scenario with custom certificates is not tested in CI.
This change is updating /functional/basic-attestation-on-localhost to generate and use custom certificates.


kkaarreell:
All test failures are expected.

PR #77, opened 2022-04-05
Fix code coverage measurement for async code

kkaarreell:
We need to set --concurrency parameter accordingly and follow
https://coverage.readthedocs.io/en/stable/cmd.html?highlight=concurrency#cmd-run


kkaarreell:
All tests failures on distribution code are expected.

PR #80, opened 2022-04-07
Use default configuration when testing distribution packages

sergio-correia:
Do not perform special setup when testing the distribution packages
with functional/basic-attestation-with-unpriviledged-agent.

From 6.3.2, the agent should run with keylime user and keylime group.

### DIFF
--- a/functional/basic-attestation-with-unpriviledged-agent/test.sh
+++ b/functional/basic-attestation-with-unpriviledged-agent/test.sh
@@ -2,10 +2,20 @@
 
 . /usr/share/beakerlib/beakerlib.sh || exit 1
 
+# Upstream.
 AGENT_ID="d432fbb3-d2f1-4a97-9ef7-75bd81c00000"
 AGENT_USER=kagent
 AGENT_GROUP=tss
 AGENT_WORKDIR=/var/lib/keylime-agent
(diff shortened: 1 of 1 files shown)


sergio-correia:
Does something like this make sense, @kkaarreell?

kkaarreell:
> Does something like this make sense, @kkaarreell?

In fact current keylime upstream code is already using `keylime` user for all tests. The respective setup is done here
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/setup/install_upstream_keylime/test.sh#L70

So this particular test had no big benefit for upstream code, I kept it running just to check a scenario where agent is using other user than keylime and different `AGENT_WORKDIR`.

For distribution packages, this was the only test running agent as non-root user, other services and other tests were running as root. Once Fedora packages adopt new default of using `keylime` user (or we do an extra setup enabling it), in theory no changes are needed in our keylime-tests. And this particular test would have same "use" as it currently has for upstream code. We can disable it eventually but I think it adds some small value to our test coverage.

kkaarreell:
I am closing this PR, the test has been fixed with https://github.com/RedHat-SP-Security/keylime-tests/commit/c452851864e96b1dc470b5827db90854861a5a7c

PR #90, opened 2022-04-28
Add test for keylime agent without mTLS

ansasaki:
Add function/basic-attestation-without-mtls test case.
The test case runs a basic attestation using the agent with mTLS disabled.

### DIFF
--- a/functional/basic-attestation-without-mtls/test.sh
+++ b/functional/basic-attestation-without-mtls/test.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+AGENT_ID="d432fbb3-d2f1-4a97-9ef7-75bd81c00000"
+
+rlJournalStart
+
+    rlPhaseStartSetup "Do the keylime setup"
+        rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
+        rlAssertRpm keylime
+        limeBackupConfig
+        # update /etc/keylime.conf
+        rlRun "limeUpdateConf tenant require_ek_cert False"
+        rlRun "limeUpdateConf cloud_agent mtls_cert_enabled False"
+        rlRun "limeUpdateConf cloud_agent enable_insecure_payload False"
+        rlRun "limeUpdateConf cloud_verifier agent_mtls_cert_enabled False"
+        # if TPM emulator is present
+        if limeTPMEmulated; then
+            # start tpm emulator
+            rlRun "limeStartTPMEmulator"
+            rlRun "limeWaitForTPMEmulator"
+            # make sure tpm2-abrmd is running
+            rlServiceStart tpm2-abrmd
+            sleep 5
+            # start ima emulator
+            export TPM2TOOLS_TCTI=tabrmd:bus_name=com.intel.tss2.Tabrmd
+            rlRun "limeInstallIMAConfig"
+            rlRun "limeStartIMAEmulator"
+        else
+            rlServiceStart tpm2-abrmd
+        fi
+        sleep 5
+        # start keylime_verifier
+        rlRun "limeStartVerifier"
+        rlRun "limeWaitForVerifier"
+        rlRun "limeStartRegistrar"
+        rlRun "limeWaitForRegistrar"
+        # when using unit files we need to adjust them
+        if [ -f /usr/lib/systemd/system/keylime_agent.service -o -f /etc/systemd/system/keylime_agent.service ]; then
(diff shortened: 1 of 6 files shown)


kkaarreell:
Hi @ansasaki , do you know if mTLS comes into play when revocation scripts from payload comes into play? If not, we could simplify the test a bit by not using payload (and do not need expect).

functional/basic-attestation-without-mtls/test.sh
```diff
@@ -0,0 +1,113 @@
...
+            # start tpm emulator
+            rlRun "limeStartTPMEmulator"
+            rlRun "limeWaitForTPMEmulator"
+            # make sure tpm2-abrmd is running
+            rlServiceStart tpm2-abrmd
+            sleep 5
+            # start ima emulator
+            export TPM2TOOLS_TCTI=tabrmd:bus_name=com.intel.tss2.Tabrmd
+            # workaround for https://github.com/keylime/rust-keylime/pull/286
+            export PATH=/usr/bin:$PATH
```
kkaarreell, line 29: Please remove this workaround, I believe it is not needed anymore.

ansasaki:
> Hi @ansasaki , do you know if mTLS comes into play when revocation scripts from payload comes into play? If not, we could simplify the test a bit by not using payload (and do not need expect).

For now, the execution of revocation scripts are not affected by mTLS disablement, but the initial payload script execution is restricted. Note that usually the revocation actions are configured/controlled by the initial payload script, so it is indirectly affected.

When mTLS is disabled, the initial payload script is executed only if `enable_insecure_payload` is set as `True`. I want to exercise this restriction, but probably I need to change the test to not set `enable_insecure_payload` and expect the payload to not be executed, then enable it and expect the payload to be executed.

The python agent is missing the [enhancement 55](https://github.com/keylime/enhancements/blob/master/55_revocation_actions_without_python.md), which adds the `allow_payload_revocation_actions` option to control whether the payload revocation actions should be executed or only pre-installed actions are enabled. This behavior should be tested in a separate test executed only for the rust agent (probably the same used to test if non-python scripts).

functional/basic-attestation-without-mtls/test.sh
ansasaki: Will do.

kkaarreell:
This is weird. CentOS Stream 8 test should be passing as it is running with upstream bits but it fails in a same way as Fedora runs with distribution rpms. @ansasaki Any idea why?

ansasaki:
> This is weird. CentOS Stream 8 test should be passing as it is running with upstream bits but it fails in a same way as Fedora runs with distribution rpms. @ansasaki Any idea why?

This is weird indeed, for me the tests should be passing on both upstream and Fedora since the exercised changes are already there. I'll investigate.

kkaarreell:
OK. Thank you.
Also, could you please add a bit more detailed description to `main.fmf` file? Something similar to what is currently being added in https://github.com/RedHat-SP-Security/keylime-tests/pull/92/files

functional/basic-attestation-without-mtls/test.sh
```diff
@@ -0,0 +1,147 @@
...
+        rlRun "expect add.expect"
+        rlRun "limeWaitForAgentStatus $AGENT_ID 'Get Quote'"
+        rlRun -s "lime_keylime_tenant -c cvlist"
+        rlAssertGrep "{'code': 200, 'status': 'Success', 'results': {'uuids':.*'$AGENT_ID'" $rlRun_LOG -E
+        rlAssertGrep "mTLS disabled" $(limeAgentLogfile)
+        rlAssertGrep "payloads cannot be deployed" $(limeAgentLogfile)
+    rlPhaseEnd
+
+    rlPhaseStartTest "Enable insecure payload and set script_payload"
+        rlRun -s "lime_keylime_tenant -v 127.0.0.1 -t 127.0.0.1 -u $AGENT_ID -c delete"
```
kkaarreell, line 89: Adding `lime_keylime_tenant -r 127.0.0.1 -t 127.0.0.1 -u d432fbb3-d2f1-4a97-9ef7-75bd81c00000 -c regdelete` below fixes the CI failure.

ansasaki: Rebased on top of current `main` and added this.

ansasaki:
> OK. Thank you. Also, could you please add a bit more detailed description to `main.fmf` file? Something similar to what is currently being added in https://github.com/RedHat-SP-Security/keylime-tests/pull/92/files

Done!

kkaarreell:
I think I will disable coverage measurement on CentOS Stream 8, this particular error already appeared in the past on this platform.
Thank you for the test. Please remember that in order to enable it in upstream keylime/keylime project one has to add the test to the packit test plan.

PR #92, opened 2022-04-30
Update fmf files and add description for each test.

Koncpa:
The fmf. files lack a more detailed description of the tests to
understand the correct purpose of each test. Add to test descriptions
the purpose of the test and the main steps that are performed in the test.


sopos:
I'm not really sure it the description in form of a list (yaml format) will be processed in an expected way.
Have  you checked `tmt tests lint` and `tmt tests show` whether the output is expected?

Koncpa:
Hi Dalibor,

I have checked `tmt tests show`, this command show also description, but when a checked `tmt tests lint` I saw generally this output on every test:

`/functional/keylime_tenant-commands-on-localhost
pass test script must be defined
pass directory path must be absolute
pass directory path must exist
pass correct attributes are used
`
Is it proper output of this command?

P.

Koncpa:
Now, I'm checked also version where isn't text in form of the list and `tmt tests lint` has same output.

Multihost/basic-attestation/main.fmf
```diff
@@ -1,4 +1,12 @@
 summary: Tests basic keylime attestation scenario on multiple hosts
+description:
```
sopos, line 2: Something like this should be the proper format of the description as it should be a text (https://tmt.readthedocs.io/en/stable/spec/tests.html#description)
```suggestion
description: |
```

kkaarreell:
Hi @Koncpa , please update PR as suggested above. Thank you.

Multihost/basic-attestation/main.fmf
Koncpa: Thanks Dalibor, I tried to update changes as you suggested.

sopos (approved):
No it look s fine to me.

Note for the future: it is easier for reviewers if you push small changes in separate commit and minimize the force-pushing as the review can than be done incrementally as well. You can always squash commits right before the merge which is the preferred way.

sopos:
Actually, I thought some of those `-` as bullet points are expected so this part is rather on @kkaarreell to confirm.

PR #95, opened 2022-05-04
[WIP] Define TPM2TOOLS_TCTI and TCTI globally in /etc/profile.d

kkaarreell:
Instead of defining these variable in each test /setup/configure_tpm_emulator defines them globally in /etc/profile.d/ and extends agent unit file.

### DIFF
--- a/Multihost/basic-attestation/test.sh
+++ b/Multihost/basic-attestation/test.sh
@@ -246,7 +246,6 @@ Agent() {
             rlServiceStart tpm2-abrmd
             sleep 5
             # start ima emulator
-            export TPM2TOOLS_TCTI=tabrmd:bus_name=com.intel.tss2.Tabrmd
             limeInstallIMAConfig
             rlRun "limeStartIMAEmulator"
         else
@@ -378,7 +377,6 @@ Agent2() {
             rlServiceStart tpm2-abrmd
             sleep 5
             # start ima emulator
-            export TPM2TOOLS_TCTI=tabrmd:bus_name=com.intel.tss2.Tabrmd
             limeInstallIMAConfig
             limeStartIMAEmulator
         else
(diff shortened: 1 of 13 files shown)


Koncpa (approved):

ansasaki (approved):

kkaarreell:
@ansasaki Should I merge it? Would you rebase your PR on top of this one and update it?
I also need to solve TCTI definition for the non-abrmd case but I will leave that for another PR I think.

ansasaki:
> @ansasaki Should I merge it? Would you rebase your PR on top of this one and update it? I also need to solve TCTI definition for the non-abrmd case but I will leave that for another PR I think.

Feel free to merge, I'll rebase my PR. And I agree, making the TCTI setting flexible to allow HW testing can be done in a separate PR.

> In fact, do you think these variables should be defined differently for HW TPM? Maybe /setup/enable_tpm_emulator is not the rigth place to set them and I should use limeLib initialization instead. WDYT?

Ideally the configuration should be the same regardless of the TPM under the broker (`tpm2-abrmd`), but it is not the case yet. I believe the configuration of these variables will converge in the future and only exceptional cases will need special configuration.

About setting these on `enable_tpm_emulator`, I think it is appropriate since the values depend on the TPM used. When we introduce tests using HW TPM, we will probably need to set them differently in something like `enable_hw_tpm`. I may be wrong, but for now I think it is OK.

kkaarreell:
Hm, it seems that for the TPM dev case we were not using tpm2-abrmd at all. While we have started the service, we didn't export these variables and therefore (IMHO) abrmd hasn't been used. I have just double checked that on a system with TPM provided by QEMU. This is probably due to the support in kernel. I think we can remove tpm2-abrmd code from tests when TPM dev is present (not runtime emulated).

kkaarreell:
I will merged this PR and prepare another one to remove abrmd start as described above.

PR #96, opened 2022-05-04
Remove tpm2-abrmd usage on systems with read TPM device

kkaarreell:
When we do not start TPM emulator in a test we do not need
tpm2-abrmd since the support for real TPM device is provided
by kernel.


kkaarreell:
I have verified that tests pass when executed on a test system with TPM device.

PR #99, opened 2022-05-06
Do not use LEGACY crypto policy

Koncpa:
The official distributed crypto-policies should now work without the need to set the policy to the LEGACY state. It will be see in CI here.

### DIFF
--- a/Library/test-helpers/create_allowlist.sh
+++ b/Library/test-helpers/create_allowlist.sh
@@ -17,7 +17,7 @@ fi
 if [ "$2" != "--" ]; then
     ALGO=$2
 else
-    ALGO=sha1sum
+    ALGO=sha256sum
 fi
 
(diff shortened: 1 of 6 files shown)


Koncpa:
As you mention @kkaarreell , test FAILED in part when we check deletion of agent in log.

kkaarreell (approved):
LGTM. Thank you. Let's discuss the rest on our keylime mtg.

kkaarreell:
I have tested in Beaker that updated tests using sha256 for RPM are passing on pbd systems with virtual TPM. Also, test PR https://github.com/keylime/keylime/pull/990 proves it passes for keylime/keylime CI and https://github.com/keylime/rust-keylime/pull/375 proves it passes also for rust-keylime CI. I think it is safe to merge it

PR #102, opened 2022-05-13
setup/install_upstream_keylime: conditionalize python3-zmq install

ueno:


kkaarreell (approved):
OK. 
Today it turned out we are heaving issues to read these kind of expressions. :-) If we won't get used to it, we would probably replace them with something like `[ -z "${}" ]` and `[ -n "${}" ]`.

ueno:
@kkaarreell yeah I agree; I've filed #104 to update those conditionals.

PR #103, opened 2022-05-13
Add test basic-attestation-with-ima-signatures

kkaarreell:
Basic test scenario for IMA signatures, using a custom generated key.
This test requires TPM device to be present on a system since kernel boot and therefore we introduce new context `swtpm` which is used to disable (filter out) this test in CI.


kkaarreell:
A test passed on a RHEL9 and Fedora34 test systems with TPM provided by QEMU.

PR #104, opened 2022-05-16
Use simpler conditional to check KEYLIME_TEST_DISABLE_REVOCATION

ueno:
Also remove the leftover failure test in Multihost/basic-attestation.


kkaarreell:
Thank you.

PR #107, opened 2022-05-18
Add /functional/install-rpm-with-ima-signature

kkaarreell:
Test requires HW TPM and therefore won't be run in CI.
Also, on F36 and lower there is an rpm issue causing test failure
https://github.com/rpm-software-management/rpm/pull/1914


kkaarreell:
Test passed on a system with HW TPM for Fedora Rawhide and RHEL9.

PR #110, opened 2022-05-26
Add functional/measured-boot-swtpm-sanity

kkaarreell:


kkaarreell:
All test failures are expected and caused by old tpm2-tools version we are using in RHEL. Because of that I am not going to merge this yet, probably we should exclude it first from running on RHEL.

kkaarreell:
I have excluded RHEL arch for this test for now. All test failures are expected, these are known bugs/regressions.

PR #113, opened 2022-05-31
Set revocation_notifiers in keylime.conf

kkaarreell:
We need to set `revocation_notifiers` accordingly so that test scenario remains functional after `revocation_notifier_webhook` gets removed from `keylime.conf` in https://github.com/keylime/keylime/pull/795


kkaarreell:
Measured boot test failures are expected.

PR #114, opened 2022-05-31
Enable adding new option to keylime.conf in limeUpdateConf

kkaarreell:
Previously, we were able only to modify an option in keylime.conf if it existed in the respective section.
Now, if the option does not exist it will be added.
This may be useful for testing a backwards compatibility code.


kkaarreell:
Measured boot test failures are expected.

PR #117, opened 2022-05-31
Remove workarounds in measured boot test scenarios

kkaarreell:


kkaarreell:
Test failures in rust and distribution test plan are expected.

PR #118, opened 2022-06-06
use create_mb_refstate from keylime_sources if available

kkaarreell:


kkaarreell:
measured boot test failure on Fedora is expected.

PR #126, opened 2022-06-14
Update fmf files

Koncpa:
Add TCMS tag to fmf file in newly created tests
for exporting tests metadatas to TCMS.


kkaarreell (approved):
LGTM. Thank you.

Issue #119, opened 2022-06-06
Implement function limeGetSourceFile


kkaarreell:
Implement function (e.g. `limeGetSourceFile`) in lime/test-helpers library that would copy or download the respective location.
The motivation is to simplify download implemented e.g. in https://github.com/RedHat-SP-Security/keylime-tests/pull/118

For `limeGetSourceFile /path/to/file` the sequence could be:
 1. Copy `/usr/loca/lib/python/keylime*/path/to/file` if present
 2. Otherwise, copy `/var/tmp/keylime_sources/path/to/file` if present
 3. Otherwise, download the file from upstream master branch

kkaarreell:
Fixed in https://github.com/RedHat-SP-Security/keylime-tests/pull/135

PR #135, opened 2022-06-16
Create functions for get source files

Koncpa:
Create new function for getting path
of file and function for copy file to
actual directory. For getting path of file is used
limeGetFilePath() function.
For copy of file from default location
to specific dir is used limeSourceFile().
Also use this new functions for tests,
where is it usefull.

Fix issue #119


Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,78 @@ limeCreateTestDir() {
...
+    elif [ ${OPTION} == "--i" ]; then
+        curl -O https://raw.githubusercontent.com/keylime/keylime/master/scripts/${NAME} && chmod a+x ${NAME}
+    fi
+
+}
+
+true <<'=cut'
+=pod
+
+=head2 limeGetSourceFile
```
kkaarreell, line 1290: It should be the other way around. `limeGetFilePath [--source|-s|--installed|-i]` should only print the complete path of the found file. While the copy function should use that path to copy that file to a requested location.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,78 @@ limeCreateTestDir() {
...
+
+Returns 0 when the when getting copy of file to actual dir is succesfull.
+
+=cut
+
+limeGetSourceFile(){
+
+    NAME=$1
+    local SOURCELOCAL=$(find /usr/local/lib -name ${NAME})
+    local SOURCELIB=$(find /usr/lib/python* -name ${NAME})
```
kkaarreell, line 1312: I would use just a single `find` command such as
```
find /usr/lib/python* /usr/local/lib/python* -print 2> /dev/null | grep pluggy/hooks.py | head -1
```
This would be for the `-i` (default) parameter.
For the `-s` parameter the `find` would search in `/var/tmp/keylime_sources`.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,78 @@ limeCreateTestDir() {
 }
 
 
+true <<'=cut'
+=pod
+
+=head2 limeCopySourceFile
```
kkaarreell, line 1259: Let's call it `limeCopyKeylimeFile`, and it should accept similar options as the other function. In addition, `-d|--download` option for downloading it from upstream could be also useful.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,92 @@ limeCreateTestDir() {
...
+=pod
+
+=head2 limeCopyKeylimeFile
+
+Copy file to actual dir from specified path or download file
+from repository to actual dir.
+
+    limeCopyKeylimeFile
+
+=over
```
kkaarreell, line 1266: Please, document also command line options.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,92 @@ limeCreateTestDir() {
...
+
+limeCopyKeylimeFile(){
+
+    local OPTION=$1
+    local NAME=$2
+
+    if [ ${OPTION} == "-s" ] ||[ ${OPTION} == "--source" ]; then
+        # get filepath to the variblae and if file exist, copy file to actual dir
+        local FILEPATH=$(limeGetKeylimeFilePath -s ${NAME})
+        if grep -q ${NAME} <<< ${FILEPATH} ; then
```
kkaarreell, line 1282: Simple tests for non-empty variable is `if [ -n "${FILEPATH}" ]; then`

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,92 @@ limeCreateTestDir() {
...
+            cp ${FILEPATH} .
+        else
+            echo "File is not in installed files!"
```
kkaarreell, line 1285: This branch is for source files so "installed files" is a bit confusing. Personally, I would use simply "File ${NAME} was not found." in both cases.

kkaarreell: In fact, you can use only one branch. Same parameter is passed to `limeGetKeylimeFilePath` so there is no need to have two distinct branches. Only the download one requires a separate branch.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,92 @@ limeCreateTestDir() {
...
+    elif [ ${OPTION} == "-i" ] ||[ ${OPTION} == "--installed" ]; then
+        # get filepath to the variblae and if file exist, copy file to actual dir
+        local FILEPATH=$(limeGetKeylimeFilePath -i ${NAME})
+        if grep -q ${NAME} <<< ${FILEPATH} ; then
+            cp ${FILEPATH} .
+        else
+            echo "File is not in installed files!"
+            return 1
+        fi
+    elif [ ${OPTION} == "-d" ] || [ ${OPTION} == "--download" ]; then
```
kkaarreell, line 1297: I am not sure we should have `-d` option at all. Probably the most appropriate (matching to the former code in `measured-boot-policy-sanity/test.sh`) usage would be:
 - if found in sources, copy file from sources
 - otherwise download from upstream code
 So in the end we would have only `-i` and `-s`.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,92 @@ limeCreateTestDir() {
...
+
+=cut
+
+limeCopyKeylimeFile(){
+
+    local OPTION=$1
+    local NAME=$2
+
+    if [ ${OPTION} == "-s" ] ||[ ${OPTION} == "--source" ]; then
+        # get filepath to the variblae and if file exist, copy file to actual dir
```
kkaarreell, line 1280: typo in `variable`

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,101 @@ limeCreateTestDir() {
...
+    fi
+}
+
+true <<'=cut'
+=pod
+
+=head2 limeGetFilePath
+
+Trying to find path of keylime script file in several directories on the system.
+Filepath is then use for copy function, where is copied file to actual dir or
```
kkaarreell, line 1307: This is not important as it describes just one possible use case. Rather say that it returns file path on STDOUT if found, empty string otherwise.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,101 @@ limeCreateTestDir() {
...
+
+Find the path to the specified filename in the location of the keylime source files.
+
+=item --install
+
+Find the path to the specified filename in the location of the keylime installed files.
+
+=back
+
+Returns 0 when the when getting path of file is succesfull.
```
kkaarreell, line 1324: Rather prints found filepath to STDOUT.
Btw, current implementation won't return non-zero exit code if file is not found. If you want such a behavior, you need to change the code accordingly.

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,101 @@ limeCreateTestDir() {
...
+
+=cut
+
+limeGetFilePath(){
+
+    local OPTION=$1
+    local FILENAME=$2
+
```
kkaarreell, line 1332: Here we should have a code that checks that the proper number of parameters have been provided. The same applies to the other function.

kkaarreell (changes requested):

Library/test-helpers/lib.sh
```diff
@@ -1253,6 +1253,111 @@ limeCreateTestDir() {
...
+    fi
+
+    local FILEPATH=$(limeGetKeylimeFilepath ${OPTION} ${NAME})
+    # get filepath to the variable and if file exist, copy file to specified dir
+    if [ -n "${FILEPATH}" ]; then
+        echo "Copying ${FILEPATH} to ${DEST}"
+        cp ${FILEPATH} ${DEST}
+    else 
+        echo "Downloading ${NAME} to ${DEST}"
+        curl -O https://raw.githubusercontent.com/keylime/keylime/master/${NAME} ${DEST}
```
kkaarreell, line 1304: curl doesn't accept dest dir as a parameter. `man curl` says
```
       -O, --remote-name
              Write output to a local file named like the remote file we  get.
              (Only  the file part of the remote file is used, the path is cut
              off.)

              The file will be saved in the current working directory. If  you
              want  the  file  saved  in  a different directory, make sure you
              change the current working directory before invoking  curl  with
              this option.
```

functional/measured-boot-swtpm-sanity/test.sh
```diff
@@ -22,11 +22,7 @@ rlJournalStart
 
         # update config.py to use our fake  binary_bios_measurements
         # for the rust agent this is handled in the /setup/install_upstream_rust_keylime task
-        if rpm -q keylime-99 &> /dev/null; then
-            CONFIG=$( ls /usr/local/lib/python*/site-packages/keylime*/keylime/config.py )
-        else
-            CONFIG=$( ls /usr/lib/python*/site-packages/keylime/config.py )
-        fi
+        CONFIG=$(limeGetFilePath --install config.py)
```
kkaarreell, line 25: Wrong function name.

Koncpa: Thanks, my bad I didn't catch that.

kkaarreell (approved):
I think this is OK, also latest test results were expected. I would like to polish doc text a bit but let's do it in a new PR.
@Koncpa Do you plan any more updates?

PR #139, opened 2022-07-01
Exercise agent-local revocation

ueno:
This series was used to test https://github.com/keylime/keylime/pull/795, which has now been merged.


kkaarreell:
Thank you. I haven't noticed this PR before I have created the other one. Let me think how to approach it in the end.

kkaarreell:
Hi @ueno , I will handle this change in the other PR, because I would like to do more changes at the same time. Most importantly, I would like to test with `agent` notifier by default, not in a dedicated test plan so I can use existing (shared) test plans for testing RHEL-9 and CentOS Stream 9. Also, I would prefer to test `agent` and `zeromq` setup within one test plan so that I get complete data for code coverage measurements.

PR #140, opened 2022-07-01
Switch default notifier from zeromq to agent

kkaarreell:


kkaarreell:
I belive that distribution test plan failures are expected, since Fedora packages do not yet support agent revocation mechanism. Despite ATM `zeromq` is being used, the failure happens due to a removal of
```
rlRun "limeUpdateConf cloud_verifier revocation_notifier_webhook yes"
```

PR #145, opened 2022-07-08
Add payload with revocation scripts as scripts

kkaarreell:
Previously, only revocation scripts as Python modules have been tested. This change introduce revocation scripts as standalone scripts, adding ability to select the required script type in a test plan or automatically.


kkaarreell:
Looks OK, those failures are expected and unrelated.

PR #147, opened 2022-07-08
Add /setup/configure_python_unbuffered_for_keylime

kkaarreell:
Sometimes there are test failures caused by the absence of a log
message in a service log, however the same message can be seen
in a log at the end of a test. Looks like a buffering problem.
This setup task sets PYTHONUNBUFFERED=1 for keylime services
as described at
https://unix.stackexchange.com/questions/505146/getting-systemd-service-logs-faster-from-my-service


kkaarreell:
That didn't help. But I have a plan.

PR #150, opened 2022-07-11
Fix /var/tmp/test_payload_file handling

kkaarreell:


functional/basic-attestation-with-unpriviledged-agent/test.sh
```diff
@@ -22,6 +22,8 @@ rlJournalStart
             rlRun "limeUpdateConf cloud_verifier revocation_notifiers ''"
             rlRun "limeUpdateConf cloud_agent listen_notifications False"
         fi
+        # change /etc/keylime.conf permissions so that agent running as ${AGENT_USER} can access it
+        rlRun "chmod 644 /etc/keylime.conf"
```
sergio-correia, line 26: Don't you need to undo this change in the cleanup step?

kkaarreell: we call `limeBackupConfig` at above and `limeRestoreConfig` in the Clean up phase.

sergio-correia (approved):

PR #166, opened 2022-07-29
Specify ima_canonical_fmt kernel boot parameter

ueno:
Otherwise the template hash is calculated over fields in platform
native format, which confuses IMA emulator on big-endian systems.


kkaarreell:
Thank you!

PR #168, opened 2022-07-29
Update test to remove pgp keys

Koncpa:
Removed pgp keys created by test.
Backup keys from previous setup.


kkaarreell:
Thank you

PR #172, opened 2022-08-15
Add revocation script which run sudo cmd

Koncpa:


functional/basic-attestation-on-localhost/payload-script/sudo_script.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+sudo -v
```
kkaarreell, line 2: We need to run some command in order to gain (and verify) root priviledges, e.g. `id`.
Later, we should check the script output to verify it has been executed as `root`.

functional/basic-attestation-on-localhost/payload-module/action_list
```diff
@@ -1 +1,2 @@
 local_action_modify_payload
+sudo_script.sh
```
kkaarreell, line 2: please, add `local_action_` prefix to script name.

functional/basic-attestation-on-localhost/test.sh
```diff
@@ -15,6 +15,8 @@ rlJournalStart
 
         # update /etc/keylime.conf
         limeBackupConfig
+        rlFileBackup --clean --missing-ok /etc/sudoers
+        rlRun "sed -i '/NOPASSWD: ALL/s/^#//g' /etc/sudoers"
```
kkaarreell, line 19: Better limit it just to a `keylime` user, not `ALL`.

kkaarreell (changes requested):

functional/basic-attestation-on-localhost/payload-module/local_action_sudo_script.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+sudo -i
```
kkaarreell, line 2: Simple `sudo whoami` should work

kkaarreell:
I am still missing a test for the sudo script output. Ideally sudo will forward output to a file which would be later on checked.

functional/basic-attestation-on-localhost/payload-script/local_action_sudo_script.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+sudo -i
```
kkaarreell, line 2: Simple `sudo whoami > /some/path` should work.

functional/basic-attestation-on-localhost/payload-module/local_action_sudo_script.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
kkaarreell, line 1: This won't work, `payload-module` can only contain Python modules, not bash scripts. You need to rewrite it in a form of a Python module.

kkaarreell:
You have accidentally added .swp file to this PR.

functional/basic-attestation-on-localhost/payload-module/local_action_sudo_script.sh
Koncpa: > You have accidentally added .swp file to this PR.

Thank you for reminding me.

Koncpa: Yeah I rewrote it into form of python module.

Koncpa:
Thanks for reminding me.

functional/basic-attestation-on-localhost/test.sh
```diff
@@ -89,6 +91,7 @@ _EOF"
             cat ${HTTP_SERVER_LOG}
             rlAssertGrep '\\"type\\": \\"revocation\\", \\"ip\\": \\"127.0.0.1\\", \\"agent_id\\": \\"d432fbb3-d2f1-4a97-9ef7-75bd81c00000\\"' ${HTTP_SERVER_LOG} -i
             rlAssertNotGrep ERROR ${HTTP_SERVER_LOG} -i
+            rlAssertGrep 'root' /var/lib/keylime/sudo_script_output.txt -i
```
kkaarreell, line 94: I think you are not removing that file in CleanUp phase.

Koncpa: Now, file is removed in clean up phase.

Koncpa:
I opened [ISSUE](https://github.com/keylime/rust-keylime/issues/440), when I try to solve different parsing of scripts in action lists.

Koncpa:
I'm just think @kkaarreell  if won't be better to do brand new case related with revocation scripts? In this scenario could also set revocation script via keylime conf. What do you think?

Koncpa:
Now I add there temporary workaround solution where we changing format of revocation script in action list based by type of keylime agent.

kkaarreell:
> I'm just think @kkaarreell if won't be better to do brand new case related with revocation scripts? In this scenario could also set revocation script via keylime conf. What do you think?

I think we have discussed that. Let's use `action_list` file here to keep test case easier to understand. Regarding `keylime.conf`, for that I would create separate test focused exclusively on revocations and verifying that scripts defined in both sources (file and keylime conf) are properly merged.
I have already filed https://github.com/RedHat-SP-Security/keylime-tests/issues/180

functional/basic-attestation-on-localhost/payload-module/local_action_sudo_script.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+os.system("sudo whoami > sudo_script_output.txt")
```
kkaarreell, line 3: wouldn't it be better to use absolute file path for the script output?

functional/basic-attestation-on-localhost/payload-script/local_action_sudo_script.py
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/python3
+
+import os
+
+os.system("sudo whoami > sudo_script_output.txt")
```
kkaarreell, line 5: absolute path?

kkaarreell: anyway, why using a python script here? bash script should work well.

functional/basic-attestation-on-localhost/test.sh
```diff
@@ -27,6 +29,11 @@ rlJournalStart
         if [ -n "$KEYLIME_TEST_DISABLE_REVOCATION" ]; then
             rlRun "limeUpdateConf cloud_agent listen_notifications False"
         fi
+        # different parsing of elements in action list in python agent
```
kkaarreell, line 32: please add a comment with the link to the reported issue and prefix with FIXME

functional/basic-attestation-on-localhost/test.sh
```diff
@@ -27,6 +29,11 @@ rlJournalStart
         if [ -n "$KEYLIME_TEST_DISABLE_REVOCATION" ]; then
             rlRun "limeUpdateConf cloud_agent listen_notifications False"
         fi
+        #FIXME different parsing of elements in action list in python agent
+        if limeIsPythonAgent; then
```
kkaarreell, line 33: This function is not yet available in the main branch. Maybe we should add it there...WDYT?

Koncpa: Yes, I know it. We will wait until you add it. Or should I add it as standalone commit? @kkaarreell

kkaarreell (approved):

PR #176, opened 2022-08-19
Add keylime* Provides to keylime dummy package

kkaarreell:


setup/install_upstream_keylime/keylime.spec
```diff
@@ -4,6 +4,13 @@ Release:	1
 Summary:	Dummy package preventing keylime RPM installation
 License:	GPLv2+	
 BuildArch:  noarch
+Provides: keylime-base
```
sergio-correia, line 7: Shouldn't these be `Conflicts`, to make sure those packages will be removed, if installed?

kkaarreell: I think that both ways would work.
Those packages should be removed due to a fact that our dummy keylime package has version 99 and provides them.

PR #177, opened 2022-08-22
Omit file in code coverage, upgrade test

Koncpa:
Omit file in code coverage, which are
not really used due to obsolete python used
only on older systems.

### DIFF
--- a/setup/generate_coverage_report/test.sh
+++ b/setup/generate_coverage_report/test.sh
@@ -26,6 +26,7 @@ UPLOAD_URL=https://transfer.sh
 #export PACKIT_SOURCE_URL=https://github.com/ansasaki/keylime
 #export PACKIT_SOURCE_SHA=a79b05642bbe04af0ef0a356afd4f5af276898bb
 
+OMIT_FILES="--omit=/var/lib/keylime/secure/unzipped/*,*/keylime/backport_dataclasses.py"
 
 rlJournalStart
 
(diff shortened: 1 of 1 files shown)


Koncpa:
The code coverage will be this PR around 75%, also with file removed in [keylime](https://github.com/keylime/keylime/commit/6283bd30f3bbc8beaef19c85debfd51e43820fb9).

setup/generate_coverage_report/test.sh
```diff
@@ -26,6 +26,7 @@ UPLOAD_URL=https://transfer.sh
 #export PACKIT_SOURCE_URL=https://github.com/ansasaki/keylime
 #export PACKIT_SOURCE_SHA=a79b05642bbe04af0ef0a356afd4f5af276898bb
 
+OMIT_FILES="--omit=/var/lib/keylime/secure/unzipped/*,/usr/local/lib/python3.10/site-packages/keylime-6.4.2-py3.10.egg/keylime/backport_dataclasses.py"
```
kkaarreell, line 29: Please, use wildcards (*) here, we won't stick to a certain keylime version forever.

kkaarreell: Moreover, this filter should probably apply also to installed distribution keylime which is not in `/usr/local/`. I think filter like `*/keylime/backport_dataclasses.py` might be enough.

PR #182, opened 2022-08-24
[ARCHIVED] Modified limeUpdateConf calls according keylime PR#1068

kkaarreell:
Updates were done following
https://github.com/keylime/enhancements/blob/master/72_config_and_simplify_tls.md


kkaarreell:
@Koncpa FYI, before this PR gets merged we need to:
 - make user all tests are passing as expected and changes from PR#1068 are merged upstream
 - remove `environment` definitions from `plans/main.fmf`
 - merge all commits except the one changing default TIMEOUT value
 - make sure there are no remaining references to `/etc/keylime.conf` and `/etc/keylime-agent.conf` as these should not be necessary anymore

PR #186, opened 2022-08-26
Test scenario which check keylime selinux policy

Koncpa:
Scenario check if keylime work properly as confined
service by selinux policy and if there are no
additional AVC's.


kkaarreell:
We have decided not to merge this one.

PR #189, opened 2022-08-29
Tests compatible both with old way of configuring Rust agent and new way of configuring Python keylime

kkaarreell:


kkaarreell:
This PR will temporarily point keylime source git to Anderson's repository. We should point it back to keylime main repo once the respective PR get's merged.

PR #193, opened 2022-09-01
Add agent and faked_measured_boot_log context

kkaarreell:
functional/measured-boot-swtpm-sanity requires modified agent that reads measured boot log from a different location. For Python agent we can set it on the system but for the Rust agent we need to do it during compilation. By setting the proper context we can control if the test will be enabled or disabled in a test plan.
This is important for testing of distribution packages that were compiled without this hack.

### DIFF
--- a/functional/measured-boot-swtpm-sanity/main.fmf
+++ b/functional/measured-boot-swtpm-sanity/main.fmf
@@ -39,3 +39,9 @@ adjust:
   - when: distro == rhel-8 or distro = centos-stream-8
     enabled: false
     because: RHEL-8 has old tpm2-tools
+  - when: agent == rust and faked_measured_boot_log is not defined
+    enabled: false
+    because: For Rust agent we are not able to fake measuredboot log during runtime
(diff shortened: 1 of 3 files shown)


Koncpa (approved):

functional/measured-boot-swtpm-sanity/main.fmf
```diff
@@ -39,3 +39,9 @@ adjust:
   - when: distro == rhel-8 or distro = centos-stream-8
     enabled: false
     because: RHEL-8 has old tpm2-tools
+  - when: agent == rust and faked_measured_boot_log is not defined
+    enabled: false
+    because: For Rust agent we are not able to fake measuredboot log during runtime
```
lkatalin, line 44: Just for my understanding, why is this the case?

kkaarreell: Pls see the updated description.

lkatalin: Thanks!

Issue #180, opened 2022-08-24
RFE: Test that revocation scripts from action_list file and revocation_actions option are properly merged


kkaarreell:
Create a new test defining 2 scripts in `revocation_actions` option and 3 scripts in `action_list` file. There should be one script listed on both places. Verify that keylime properly merges these two inputs and 4 scripts in total are run (the script which is present in both lists should be run just once).

PR #194, opened 2022-09-01
WIP: Test scenario checking using multiple revocation scripts

Koncpa:
Scenario check running multiple revocation scripts via
multiple options how to run.

#180

### DIFF
--- a/functional/keylime-revocation_scripts/main.fmf
+++ b/functional/keylime-revocation_scripts/main.fmf
@@ -1,0 +1,29 @@
+summary: Tests keylime revocations scripts
+description: |
+    Running all services on localhost.
+    Starts verifier, registrar, agent.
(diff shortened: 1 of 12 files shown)


Koncpa:
I found out issue with specifying revocation script in keylime conf.

[ISSUE](https://github.com/keylime/keylime/issues/1095)

Koncpa:
This is now blocked by the missing fix for the actual keylime issue.

Koncpa:
> @Koncpa Do you know is this is still blocked on upstream fix?

The problem I opened in the upstream is still unresolved.

ansasaki:
Hi, the support for lists in `revocation_actions` option was introduced in https://github.com/keylime/rust-keylime/pull/604

The previous created issue for the python agent is no longer relevant, but I think this test can be used for the rust agent.

PR #196, opened 2022-09-06
Keep only necessary configuration files in Multihost/basic-attestation

ansasaki:
Remove the unnecessary configuration files for the other components in each machine.

Also fix the revocation configuration and simplify the TLS configuration in the agent by using the `tls_dir` option

### DIFF
--- a/Multihost/basic-attestation/test.sh
+++ b/Multihost/basic-attestation/test.sh
@@ -106,7 +106,7 @@ Verifier() {
         rlRun "limeUpdateConf verifier trusted_client_ca '[\"cacert.pem\"]'"
         rlRun "limeUpdateConf verifier server_cert verifier-cert.pem"
         rlRun "limeUpdateConf verifier server_key verifier-key.pem"
-	rlRun "limeUpdateConf verifier client_cert verifier-client-cert.pem"
+        rlRun "limeUpdateConf verifier client_cert verifier-client-cert.pem"
         rlRun "limeUpdateConf verifier client_key verifier-client-key.pem"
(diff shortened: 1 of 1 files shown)


Multihost/basic-attestation/test.sh
```diff
@@ -207,7 +217,7 @@ Agent() {
         id keylime && rlRun "chown -R keylime.keylime ${SECUREDIR}"
 
         # common configuration goes here
-        rlRun "limeUpdateConf general receive_revocation_ip ${VERIFIER_IP}"
+        rlRun "limeUpdateConf agent revocation_notification_ip ${VERIFIER_IP}"
```
kkaarreell, line 220: Could you please move this line to the agent configuration section and also remove `# common configuration goes here`

Multihost/basic-attestation/test.sh
```diff
@@ -353,13 +369,14 @@ Agent2() {
...
```
kkaarreell, line 372: Same as above

kkaarreell (changes requested):
Thank you for the update, it was in my queue for some time already...

Multihost/basic-attestation/test.sh
ansasaki: Sure, I'll fix this.

Multihost/basic-attestation/test.sh
ansasaki: Sure

ansasaki:
This PR was targeting the `new_config` branch which was merged into `main`. I'll create a new PR targeting `main`.

Issue #132, opened 2022-06-15
RFE: Test scenario with multiple ima signature verification keys


kkaarreell:
ATM we are using just a single key. We should probably extend existing test (or created a new one?) so that at least two different keys are used. Also, we may need to update `limeInstallIMAKeys` function from test-helpers library so it would generate new keys to a custom location or filename.

Koncpa:
Solution merged, closed.

PR #198, opened 2022-09-09
Update scenario to use multiple ima signature verification keys

Koncpa:
#132

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -1147,9 +1147,21 @@ true <<'=cut'
 
 =head2 limeInstallIMAKeys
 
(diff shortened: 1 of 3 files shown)


Library/test-helpers/lib.sh
```diff
@@ -1147,10 +1147,20 @@ true <<'=cut'
 
 =head2 limeInstallIMAKeys
 
-Generate and install IMA/EVM keys to /etc/keys/
+Generate and install IMA/EVM keys to /etc/keys/ or install custom keys to specified destination.
```
kkaarreell, line 1150: or to the specified destination.

kkaarreell (changes requested):

kkaarreell (approved):

PR #200, opened 2022-09-15
Add phase checking keylime_tenant --key and --payload options

Koncpa:


Koncpa:
This is now blocked by the missing fix for the actual keylime issue.

functional/keylime_tenant-commands-on-localhost/test.sh
```diff
@@ -34,13 +34,21 @@ rlJournalStart
         rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
         # create allowlist and excludelist
         limeCreateTestPolicy
+        #encrypting of message
+        rlRun "openssl enc -a -in payload/secret_message.txt -out payload/secret_encrypted.txt -e -aes256 -k payload/key.txt"
```
ansasaki, line 38: I think this is incorrect. I think this will not set the correct cipher mode to GCM and also the `-k` option seems to be used to specify a passphrase (for a PBKDF) and not the actual key. The key is also b64 encoded in the provided file, and is stored together with the `U` and `V` keys.

I suggest you to use the `keylime_userdata_encrypt` tool instead, which will generate the keys and the payload correctly.

Koncpa: @ansasaki Oh thank you for review and clarification, now it works. So it was wrong misconfiguration of encryption scheme. Interesting is, that for python agent was this approach suitable. And also good to know, that keylime have tool, which can encrypt user data.

ansasaki: Probably the python agent doesn't fail because it doesn't validate if the provided payload is base64 encoded.

Koncpa: Ok, thank you for clarification.

kkaarreell (approved):

kkaarreell:
Test failure in /functional/tenant-runtime-policy-sanity is expected as this is a known keylime upstream regression.

PR #202, opened 2022-09-20
Adapt the tests for the new rust agent configuration format

ansasaki:
Modify the tests to be compatible with the TOML format used in the rust agent. The main differences are:

- Booleans: on TOML format they are case sensitive. Accepted values are `true` and `false`
- Strings: on TOML format they need to be double quoted.

These changes should be compatible with INI format.

This branch is temporarily used for testing the PR:
https://github.com/keylime/rust-keylime/pull/449

### DIFF
--- a/functional/basic-attestation-with-unpriviledged-agent/test.sh
+++ b/functional/basic-attestation-with-unpriviledged-agent/test.sh
@@ -16,16 +16,11 @@ rlJournalStart
         rlAssertRpm keylime
         limeBackupConfig
         # update keylime conf
-        limeIsPythonAgent && AGENT_CONFIG_SECTION=agent || AGENT_CONFIG_SECTION=cloud_agent
         rlRun "limeUpdateConf tenant require_ek_cert False"
         rlRun "limeUpdateConf revocations enabled_revocation_notifications '[\"${REVOCATION_NOTIFIER}\"]'"
         if [ -n "$KEYLIME_TEST_DISABLE_REVOCATION" ]; then
             rlRun "limeUpdateConf revocations enabled_revocation_notifications '[]'"
-            if limeIsPythonAgent; then
-                rlRun "limeUpdateConf ${AGENT_CONFIG_SECTION} enable_revocation_notifications False"
-            else
-                rlRun "limeUpdateConf ${AGENT_CONFIG_SECTION} listen_notifications False"
-            fi
+            rlRun "limeUpdateConf agent enable_revocation_notifications false"
         fi
         # change /etc/keylime.conf permissions so that agent running as ${AGENT_USER} can access it
         rlRun "find /etc/keylime -type f -exec chmod 444 {} \;"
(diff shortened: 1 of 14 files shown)


ansasaki:
This made me find 2 details on python keylime agent:

- For the `ek_handle` option, it parses the handle as a number instead of a string. So, we cannot use quoted string like `ek_handle = "0x81010001"`
- For the `run_as` option, it does not strip the quotes from the string, making `run_as = "keylime:tss"` invalid. We need to use unquoted strings.

functional/basic-attestation-with-unpriviledged-agent/test.sh
```diff
@@ -50,16 +45,20 @@ rlJournalStart
         rlRun "limeStartRegistrar"
         rlRun "limeWaitForRegistrar"
         # do special configuration for the agent
-        rlRun "limeUpdateConf ${AGENT_CONFIG_SECTION} run_as ${AGENT_USER}:${AGENT_GROUP}"
+        if limeIsPythonAgent; then
+            rlRun "limeUpdateConf agent run_as ${AGENT_USER}:${AGENT_GROUP}"
+        else
+            rlRun "limeUpdateConf agent run_as '\"${AGENT_USER}:${AGENT_GROUP}\"'"
```
kkaarreell, line 51: Is this something that won't work with Python agent?

kkaarreell:
LGTM.. Though later we may need to update other tests not run in CI

functional/basic-attestation-with-unpriviledged-agent/test.sh
ansasaki: After trying to make them the same, I realized that the python agent does not strip the quotes from the strings obtained from the configuration in all cases (this being one of them)

So no, while the TOML format requires the quotes, the python agent does not work correctly when the strings are quoted.

kkaarreell: I see.

ansasaki:
I'll make more changes as Sergio requested the configuration files and snippets to be installed following the python components structures for consistency. So the rust agent configuration will be placed at `/etc/keylime/agent.conf` and the snippets at `/etc/keylime/agent.conf.d`

kkaarreell:
Since you are already updating the Rust agent install task, could you incorporate the configuration using environment variables as we do for Python?
https://github.com/RedHat-SP-Security/keylime-tests/blob/a5ad281392b1cea5d90f26c225e9996f2b9f102c/setup/install_upstream_keylime/test.sh#L7

This way, next time one could override repo on a test plan side, e.g. in a Rust agent PR only.

ansasaki:
> Since you are already updating the Rust agent install task, could you incorporate the configuration using environment variables as we do for Python?
> 
> https://github.com/RedHat-SP-Security/keylime-tests/blob/a5ad281392b1cea5d90f26c225e9996f2b9f102c/setup/install_upstream_keylime/test.sh#L7
> 
> This way, next time one could override repo on a test plan side, e.g. in a Rust agent PR only.

Sure, I'll include this change

PR #204, opened 2022-09-21
Remove sudo script checking

Koncpa:


kkaarreell:
I was thinking whether commenting out relevant parts wouldn't be enough, but maybe better to remove it and store link to this PR somewhere so it is available if we would like to restore it.
Thank you.

Koncpa:
I had stored branch with PR, so in future we can restore it.

PR #213, opened 2022-10-18
Draft: Do not run upstream tests as root

ansasaki:
The goal is to make the tests compatible with https://github.com/keylime/keylime/pull/1132 in which the upstream tests are not required to be executed as root anymore.

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -2,3 +2,8 @@ context:
   swtpm: yes
   agent: python
   faked_measured_boot_log: true
+
+environment:
+  KEYLIME_UPSTREAM_URL: "https://github.com/ansasaki/keylime.git"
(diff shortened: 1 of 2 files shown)


kkaarreell:
HI @ansasaki , are those missing dependencies something that has been newly added to keylime? Or maybe a package set installed by default on a test system has changed.

ansasaki:
> HI @ansasaki , are those missing dependencies something that has been newly added to keylime? Or maybe a package set installed by default on a test system has changed.

There are no new dependencies, it is just that the upstream test suite were not installing the dependencies needed to build the rust agent, but now they are being installed.

> EDIT: It rather looks like a bug in a test case code

Yes, this is not finished yet.

PR #217, opened 2022-11-01
Measured code coverage for rust

Koncpa:
Pull request is still in progress and blocked due [Graceful shutdown of the agent](https://github.com/keylime/rust-keylime/issues/363) issue in keylime rust agent upstream.

### DIFF
--- a/plans/rust-keylime-tests.fmf
+++ b/plans/rust-keylime-tests.fmf
@@ -3,6 +3,7 @@ summary:
 
 environment+:
   TPM_BINARY_MEASUREMENTS: /var/tmp/binary_bios_measurements
+  KEYLIME_RUST_CODE_COVERAGE: 1
 
 context+:
   agent: rust
@@ -32,6 +33,13 @@ adjust:
         order: 99
         script:
           - yum -y downgrade tpm2-tss
+  # disable code coverage measurement everywhere except F37 and CS9
+  - when: distro != centos-stream-9 and distro != fedora-37
+    environment+:
+       KEYLIME_RUST_CODE_COVERAGE: 0
+    discover+:
+       test-:
+         - /setup/generate_usptream_rust_keylime_code_coverage
 
 discover:
   how: fmf
@@ -42,6 +50,8 @@ discover:
    - /setup/enable_keylime_debug_messages
    - /setup/configure_kernel_ima_module/ima_policy_signing
    - "/functional/.*"
+   - /upstream/run_rust_keylime_tests
+   - /setup/generate_usptream_rust_keylime_code_coverage
 
 execute:
     how: tmt
--- a/setup/generate_usptream_rust_keylime_code_coverage/main.fmf
+++ b/setup/generate_usptream_rust_keylime_code_coverage/main.fmf
@@ -1,0 +1,11 @@
+summary: Process gathered code coverage data for the rust agent and generate a report
+description:
+contact: Patrik Koncity <pkoncity@redhat.com>
+component:
+- keylime
+test: ./test.sh
+tag:
+ - setup
+framework: beakerlib
+duration: 10m
+enabled: true
--- a/setup/generate_usptream_rust_keylime_code_coverage/test.sh
+++ b/setup/generate_usptream_rust_keylime_code_coverage/test.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+# vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+
+UPLOAD_URL=https://transfer.sh
+
+
+rlJournalStart
+
+    rlPhaseStartSetup "Collect code coverage for rust components"
+        rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
+        #delete coverage script, in dir are mandatory only coverage files
+        rlRun "pushd ${__INTERNAL_limeCoverageDir}"
+        # export in format for user, gather code cover for all rust binaries
+        rlRun "grcov . --binary-path /usr/local/bin/ -s . -t html --branch --ignore-not-existing -o e2e_coverage.html"
+        # export coverage report in format compatible for codecov, gather code cover for all rust binaries
+        rlRun "grcov . --binary-path /usr/local/bin/ -s . -t lcov --branch --ignore-not-existing -o e2e_coverage.txt"
+        # create tar file for uploading coverage file
+        rlRun "tar --create --file e2e_coverage.tar e2e_coverage.html"
+        rlFileSubmit  e2e_coverage.tar
+        rlFileSubmit e2e_coverage.txt
+        if [ -f "e2e_coverage.tar" ]; then
+            # upload e2e report in tar.gz
+            rlRun -s "curl --upload-file e2e_coverage.tar $UPLOAD_URL"
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "HTML code coverage report is available as GZIP archive at $URL"
+        fi
+        if [ -f "e2e_coverage.txt" ]; then
+            #upload e2e report in .txt format
+            rlRun -s "curl --upload-file e2e_coverage.txt $UPLOAD_URL"
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "e2e_coverage.txt report is available at $URL"
+        fi
+        if [ -f "upstream_coverage.tar" ]; then
+            #upload upstream report in .tar
+            rlRun -s "curl --upload-file upstream_coverage.tar $UPLOAD_URL"
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "HTML code coverage report is available as GZIP archive at $URL"
+            fi
+        if [ -f "upstream_coverage.xml" ]; then
+            #upload
+            rlRun -s "curl --upload-file upstream_coverage.xml $UPLOAD_URL"
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "upstream_coverage.xml report is available at $URL"
+        fi
+        rlRun "popd"
+    rlPhaseEnd
+
+rlJournalEnd
+
--- a/setup/install_upstream_rust_keylime/test.sh
+++ b/setup/install_upstream_rust_keylime/test.sh
@@ -10,6 +10,7 @@
 rlJournalStart
 
     rlPhaseStartSetup "Build and install rust-keylime bits"
+        rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
         if [ -d /var/tmp/rust-keylime_sources ]; then
             rlLogInfo "Compiling rust-keylime bits from /var/tmp/rust-keylime_sources"
         else
@@ -25,11 +26,25 @@ rlJournalStart
                 [ -f ${FILE} ] && rlRun "sed -i 's%/sys/kernel/security/tpm0/binary_bios_measurements%${TPM_BINARY_MEASUREMENTS}%' $FILE"
             done
         fi
+
+        if [ "${KEYLIME_RUST_CODE_COVERAGE}" == "1" -o "${KEYLIME_RUST_CODE_COVERAGE}" == "true" ]; then
+            rlRun "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y"
+            rlRun "source \"$HOME/.cargo/env\""
+            rlRun "rustup default nightly"
+            rlRun "rustup component add llvm-tools-preview"
+            sleep 3
+            #install parser for code coverage files
+            rlRun "cargo install grcov"
+            # -Z is deprecated, use -C
+            rlRun "export RUSTFLAGS='-Cinstrument-coverage'"
+        fi
+        #build
         rlRun "cargo build"
+
         rlAssertExists target/debug/keylime_agent
         [ -f /usr/local/bin/keylime_agent ] && rlRun "mv /usr/local/bin/keylime_agent /usr/local/bin/keylime_agent.backup"
         rlRun "cp target/debug/keylime_agent /usr/local/bin/keylime_agent"
-        if [ -n "${RUST_IMA_EMULATOR}" ]; then
+        if [ -n "${RUST_IMA_EMULATOR}" ] || [ -n "${KEYLIME_RUST_CODE_COVERAGE}" ]; then
             rlRun "cp target/debug/keylime_ima_emulator /usr/local/bin/keylime_ima_emulator"
         fi
         if [ -f keylime-agent.conf ]; then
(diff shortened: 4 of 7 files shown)


plans/rust-keylime-tests.fmf
```diff
@@ -29,6 +29,7 @@ discover:
    - /setup/install_upstream_rust_keylime
    - /setup/configure_kernel_ima_module/ima_policy_signing
    - "/functional/.*"
+   - /setup/generate_usptream_keylime_code_coverage
```
kkaarreell, line 32: please name it `/setup/generate_usptream_rust_keylime_code_coverage`

setup/generate_usptream_keylime_code_coverage/test.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+# define RUST_IMA_EMULATOR variable to install also rust IMA emulator
+
+rlJournalStart
+
+    rlPhaseStartSetup "Build and install rust-keylime bits"
```
kkaarreell, line 9: either the phase name is confusing or you are doing something different than advertized

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -23,7 +23,26 @@ rlJournalStart
...
+
+        rlRun "dnf remove -y rust"
+        rlRun "curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain none -y"
+        rlRun "source '$HOME/.cargo/env'"
+        rlRun "rustup default nightly"
+        rlRun "rustup component add llvm-tools-preview"
+        sleep 3
+        #install stuff for code coverage
+        rlRun "cargo install grcov"
+        rlRun "export RUSTFLAGS='-Cinstrument-coverage'"
```
kkaarreell, line 40: to make it work also for systemd services you should define these environment in a unit file

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -53,16 +72,32 @@ _EOF'
...
+# set variable containing name of the currently running test
+Environment=\"LLVM_PROFILE_FILE=/var/tmp/limeLib/coverage/rust_keylime_codecoverage-%p-%m.profraw\"
+# we need to change WorkingDirectory since .profraw* files will be stored there
+WorkingDirectory=/var/tmp/limeLib/coverage
+_EOF"
+        rlRun "systemctl daemon-reload"
 
     rlPhaseStartTest "Test installed binaries"
         rlRun "keylime_agent --help" 0,1
+        bash
```
kkaarreell, line 99: This `bash` here will break your CI jobs.

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -53,16 +72,32 @@ _EOF'
...
         fi
-    rlPhaseEnd
+
+        id keylime && rlRun "chown -R keylime /var/tmp/limeLib && chmod -R g+w /var/tmp/limeLib"
+        grep -q LLVM_PROFILE_FILE /etc/bashrc || rlRun "echo 'export LLVM_PROFILE_FILE=/var/tmp/limeLib/coverage/rust_keylime_codecoverage-%p-%m.profraw' >> /etc/bashrc"
+            if [ $DIREXIST == "FALSE" ];then
+            rlRun "mkdir -p /etc/systemd/system/keylime_agent.service.d"
+            fi
+            rlRun "cat > /etc/systemd/system/keylime_agent.service.d/10-coverage.conf <<_EOF
+[Service]
```
kkaarreell, line 89: Not sure if Rust allows it but what we do for Python is that we define a unique context for every single test.. This way we can measure code coverage for individual tests. But I guess we can implement it later eventually, no need to rush.
Just for the record, relevant code pieces are in
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/Library/test-helpers/lib.sh#L1711
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/setup/enable_keylime_coverage/test.sh#L30

Koncpa: Sure, I'll look more to this area after resolving simple code coverage of rust components.

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -7,6 +7,9 @@
 [ -n "${RUST_KEYLIME_UPSTREAM_URL}" ] || RUST_KEYLIME_UPSTREAM_URL="https://github.com/keylime/rust-keylime.git"
 [ -n "${RUST_KEYLIME_UPSTREAM_BRANCH}" ] || RUST_KEYLIME_UPSTREAM_BRANCH="master"
 
+
+COVERAGE_DIR='/var/tmp/limeLib/coverage/'
```
kkaarreell, line 11: Do not define the variable but rather use `$__INTERNAL_limeCoverageDir` defined by `lime`library.  Yeah, we should probably strip the `__INTERNAL_` prefix.

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -23,7 +26,21 @@ rlJournalStart
             rlRun "sed -i 's%/sys/kernel/security/tpm0/binary_bios_measurements%${TPM_BINARY_MEASUREMENTS}%' src/common.rs"
             rlRun "sed -i 's%/sys/kernel/security/tpm0/binary_bios_measurements%${TPM_BINARY_MEASUREMENTS}%' src/main.rs"
         fi
+
+        if [ -z "$KEYLIME_RUST_CODE_COVERAGE" ]; then
```
kkaarreell, line 30: Hi @ansasaki , could you please take a look at this code block? I am wondering if all added lines are necessary or whether some could possibly duplicate what is already configured in a rust-keylime repository.

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -57,6 +74,15 @@ _EOF'
 [Service]
 Environment=\"RUST_LOG=keylime_agent=trace\"
 _EOF"
+            if [ -z "$KEYLIME_RUST_CODE_COVERAGE" ]; then
+                rlRun "cat > /etc/systemd/system/keylime_agent.service.d/10-coverage.conf <<_EOF
```
kkaarreell, line 78: I am thinking we should maybe rename the file to `15-rust-coverage.conf` to avoid conflict with the one created for Python. It should not do any harm if both files would be present, however with conflicting names in place the order of setup script execution would matter.

setup/install_upstream_rust_keylime/test.sh
ansasaki: Unfortunately, this is still necessary as it is not possible to set the measured boot log location via configuration or environment variable. We should probably add an option to allow setting the path.

plans/rust-keylime-tests.fmf
```diff
@@ -31,6 +32,8 @@ discover:
    - /setup/enable_keylime_debug_messages
    - /setup/configure_kernel_ima_module/ima_policy_signing
    - "/functional/.*"
+   - /setup/generate_usptream_rust_keylime_code_coverage
```
kkaarreell, line 35: typo in "upstream"

plans/rust-keylime-tests.fmf
```diff
@@ -3,6 +3,7 @@ summary:
 
 environment+:
   TPM_BINARY_MEASUREMENTS: /var/tmp/binary_bios_measurements
+  KEYLIME_RUST_CODE_COVERAGE: 1
```
kkaarreell, line 6: We should make it conditional as for Python and measure coverage only on Fedora37 and CS9.

plans/rust-keylime-tests.fmf
```diff
@@ -41,7 +42,11 @@ discover:
    - /setup/install_upstream_rust_keylime
    - /setup/enable_keylime_debug_messages
    - /setup/configure_kernel_ima_module/ima_policy_signing
-   - "/functional/.*"
+   - /functional/basic-attestation-on-localhost
+#   - "/functional/.*"
+   - /upstream/run_rust_keylime_tests
+   - /setup/generate_usptream_rust_keylime_code_coverage
```
kkaarreell, line 48: also this should be conditionally removed (in `adjust`).

setup/generate_usptream_rust_keylime_code_coverage/main.fmf
```diff
@@ -0,0 +1,11 @@
+summary: Generate codecoverage for rust agent
```
kkaarreell, line 1: Process gathered code coverage data for the rust agent and generate a report

setup/generate_usptream_rust_keylime_code_coverage/test.sh
```diff
@@ -0,0 +1,46 @@
...
+        if [ "${UPLOAD_COVERAGE}" == "1" ]; then
+            # upload coverage.tar.gz
+            rlRun -s "curl --upload-file rust_coverage.tar $UPLOAD_URL"
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "HTML code coverage report is available as GZIP archive at $URL"
+            #upload coverage_rust.txt reports
+            rlRun -s "curl --upload-file rust_coverage.txt $UPLOAD_URL"
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "rust_coverage.txt report is available at $URL"
+            rlRun -s "curl --upload-file tarpaulin-report.tar $UPLOAD_URL"
```
kkaarreell, line 33: it would be nice if the `tarpauling-report` file had a different name that better describe which tests were run when during the measurement. Please use names like . `e2e-tests` and `upstream-testsuite` or similar.

setup/generate_usptream_rust_keylime_code_coverage/test.sh
```diff
@@ -0,0 +1,46 @@
...
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "HTML code coverage report is available as GZIP archive at $URL"
+            #upload coverage_rust.txt reports
+            rlRun -s "curl --upload-file cobertura.xml $UPLOAD_URL"
```
kkaarreell, line 37: same as above

setup/generate_usptream_rust_keylime_code_coverage/test.sh
```diff
@@ -0,0 +1,46 @@
...
+            URL=$( grep -o 'https:[^"]*' $rlRun_LOG )
+            rlLogInfo "cobertura.xml report is available at $URL"
+        fi
+        rlRun "rm -rf ${__INTERNAL_limeCoverageDir}/*"
```
kkaarreell, line 41: Why do you delete the content here?

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -25,11 +26,26 @@ rlJournalStart
                 [ -f ${FILE} ] && rlRun "sed -i 's%/sys/kernel/security/tpm0/binary_bios_measurements%${TPM_BINARY_MEASUREMENTS}%' $FILE"
             done
         fi
+
+        if [ -n "$KEYLIME_RUST_CODE_COVERAGE" ]; then
```
kkaarreell, line 30: I think better to use
```
[ "$KEYLIME_RUST_CODE_COVERAGE" == "1" -o "$KEYLIME_RUST_CODE_COVERAGE" == "true" ]
```
so we can be possibly explicit in a test plan stating `KEYLIME_RUST_CODE_COVERAGE: false` or similar.

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -59,6 +75,29 @@ _EOF'
 [Service]
 Environment=\"RUST_LOG=keylime_agent=trace\"
 _EOF"
+            if [ -n "$KEYLIME_RUST_CODE_COVERAGE" ]; then
```
kkaarreell, line 78: same as above. Maybe process that variable at the beginning of the script like
```
if [ "$KEYLIME_RUST_CODE_COVERAGE" == "1" -o "$KEYLIME_RUST_CODE_COVERAGE" == "true" ]; then
  MEASURE_COVERAGE=true
else
  MEASURE_COVERAGE=false
fi
```
and later use just `if $MEASURE_COVERAGE; then ...`

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -59,6 +75,29 @@ _EOF'
...
+                rlRun 'cat > ${__INTERNAL_limeCoverageDir}/coverage-script-stop.sh <<_EOF
+#!/bin/sh
+pushd ${__INTERNAL_limeCoverageDir}
+COV_FILE=\$(mktemp rust_keylime_codecoverage-XXXXX  --suffix=.profraw)
+cp rust_keylime_codecoverage.profraw \${COV_FILE}
+popd
+_EOF'
+                rlRun "cat > /etc/systemd/system/keylime_agent.service.d/15-coverage.conf <<_EOF
+[Service]
+#ExecStartPre=sh ${__INTERNAL_limeCoverageDir}/coverage-script-start.sh
```
kkaarreell, line 91: Is this needed or now? if not, pls remove it.

upstream/run_rust_keylime_tests/main.fmf
```diff
@@ -0,0 +1,14 @@
+summary: Run tests from keylime rust upstream testsuite and gather coverage
```
kkaarreell, line 1: ... and eventually gather code coverage

upstream/run_rust_keylime_tests/test.sh
```diff
@@ -0,0 +1,53 @@
...
+            # start ima emulator
+            rlRun "limeInstallIMAConfig"
+            rlRun "limeStartIMAEmulator"
+        fi
+        rlRun "pushd /var/tmp/rust-keylime_sources"
+        rlRun "cargo install cargo-tarpaulin"
+    rlPhaseEnd
+
+    rlPhaseStartTest "Run unit tests"
+        #run cargo tarpaulin code coverage
```
kkaarreell, line 36: I think we should make all the code coverage related commands conditional.. so it can also work without a code coverage measurement. For troubleshooting we may not want to measure code coverage.

kkaarreell (changes requested):

setup/generate_usptream_rust_keylime_code_coverage/test.sh
Koncpa: When you run plan several times, you have in coverage dir a files from previous runs. So I decided to after each run wipe out.

kkaarreell: Yep. But TBH, I don't expect us to generate multiple different coverage reports in sequence. I think it is better to keep it in place in case you would like to work with it manually.
Also, with this approach you would delete also Python coverage data that are stored in the very same directory.

setup/install_upstream_rust_keylime/test.sh
Koncpa: Why we would use another variable, when we can you variable already defined in plan? @kkaarreell

upstream/run_rust_keylime_tests/test.sh
Koncpa: I tried to make conditional, so please take a look if it's sufficient.

setup/install_upstream_rust_keylime/test.sh
Koncpa: So I used the variable from plan if don't mind.

setup/generate_usptream_rust_keylime_code_coverage/test.sh
```diff
@@ -0,0 +1,44 @@
...
+rlJournalStart
+
+    rlPhaseStartSetup "Collect code coverage for rust components"
+        rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
+        #delete coverage script, in dir are mandatory only coverage files
+        rlRun "pushd ${__INTERNAL_limeCoverageDir}"
+        # export in format for user
+        rlRun "grcov . --binary-path /usr/local/bin/keylime_agent -s . -t html --branch --ignore-not-existing -o e2e_coverage.html"
+        # export coverage report in format compatible for codecov
+        rlRun "grcov . --binary-path /usr/local/bin/keylime_agent -s . -t lcov --branch --ignore-not-existing -o e2e_testsuite.txt"
```
kkaarreell, line 19: I am confused. If these are the same data just with a different format why don't you call it e2e_coverage.txt?

setup/generate_usptream_rust_keylime_code_coverage/test.sh
```diff
@@ -0,0 +1,44 @@
...
+        # create tar file for uploading coverage file
+        rlRun "tar --create --file e2e_testsuite.tar e2e_coverage.html"
+        rlFileSubmit  e2e_testsuite.tar
+        rlFileSubmit e2e_testsuite.txt
+        if [ "${UPLOAD_COVERAGE}" == "1" ]; then
+            # upload coverage.tar.gz
+            rlRun -s "curl --upload-file e2e_testsuite.tar $UPLOAD_URL"
```
kkaarreell, line 26: would it make sense to upload individual files only when they really exist? `if [ -f FILE ]; then ...`

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -25,11 +26,25 @@ rlJournalStart
...
+            rlRun "export RUSTFLAGS='-Cinstrument-coverage'"
+        fi
+        #build
         rlRun "cargo build"
+
         rlAssertExists target/debug/keylime_agent
         [ -f /usr/local/bin/keylime_agent ] && rlRun "mv /usr/local/bin/keylime_agent /usr/local/bin/keylime_agent.backup"
         rlRun "cp target/debug/keylime_agent /usr/local/bin/keylime_agent"
-        if [ -n "${RUST_IMA_EMULATOR}" ]; then
+        if [ -n "${RUST_IMA_EMULATOR}" ] || [ -n "$KEYLIME_RUST_CODE_COVERAGE" ]; then
```
kkaarreell, line 47: I get the point, still not sure if it is really needed to modify this condition.

upstream/run_rust_keylime_tests/main.fmf
```diff
@@ -0,0 +1,14 @@
+summary: Run tests from keylime rust upstream testsuite and gather coverage and eventually gather code coverage
```
kkaarreell, line 1: remove the first "and gather coverage" as it is there twice.

kkaarreell (changes requested):
Only few nitpicks, otherwise looks good. Thank you!.

kkaarreell (approved):

PR #227, opened 2022-11-04
Enable distribution plan for C9S

kkaarreell:
C9S already has keylime with the new syntax so we can run the distribution test plan there.


Koncpa (approved):

kkaarreell:
Dropping, I didn't realize that c9s contain just the Rust agent.

PR #262, opened 2022-12-02
Make --install a default in limeCopyKeylimeFile

kkaarreell:


kkaarreell:
@Koncpa I have noticed that we are not shipping `create_mb_refstate` script and therefore I left existing tests untouched.

kkaarreell:
> @Koncpa I have noticed that we are not shipping `create_mb_refstate` script and therefore I left existing tests untouched.

Update, wit ship it in `/usr/share/keylime` but upstream installation script does not install it. Maybe we should modify our keylime install task to copy it to /usr/share/keylime so we can always find it there. WDYT?

Koncpa:
> > @Koncpa I have noticed that we are not shipping `create_mb_refstate` script and therefore I left existing tests untouched.
> 
> Update, wit ship it in `/usr/share/keylime` but upstream installation script does not install it. Maybe we should modify our keylime install task to copy it to /usr/share/keylime so we can always find it there. WDYT?

Yeah, sure...will be the most simple. Do you will implement it in this PR?

Koncpa (approved):

kkaarreell:
> Yeah, sure @kkaarreell ...it will be the most simple. Do you will implement it in this PR?

Yes, I have just pushed an update.

PR #272, opened 2022-12-08
Prepare reproducer for keylime PR#486

kkaarreell:


THS-on:
@kkaarreell this does not point the my PR, correct? For this there should also be a change to https://github.com/RedHat-SP-Security/keylime-tests/blob/main/setup/install_upstream_rust_keylime/test.sh

kkaarreell:
> @kkaarreell this does not point the my PR, correct? For this there should also be a change to https://github.com/RedHat-SP-Security/keylime-tests/blob/main/setup/install_upstream_rust_keylime/test.sh

You are right! I did the change locally but forgot to commit it. Let me fix it right now.

kkaarreell:
Closing as this is no longer needed.

PR #274, opened 2022-12-09
Add support for AVC_CHECK_AUSEARCH_PARAMS

kkaarreell:


kkaarreell:
@Koncpa @sergio-correia I was thinking, maybe for Rawhide keylime CI we do not need to see (and fail on) all AVCs but only on keylime related ones. We could define `AVC_CHECK_AUSEARCH_PARAMS="-se keylime"` in a test plan for Rawhide. WDYT?

sergio-correia:
@kkaarreell: +1 from me, I think this makes sense.

Koncpa (approved):

PR #279, opened 2022-12-14
keylime install - use less pkgs from custom repos

kkaarreell:


setup/install_upstream_keylime/test.sh
```diff
@@ -32,20 +32,21 @@ gpgkey=https://download.copr.fedorainfracloud.org/results/scorreia/keylime/pubke
...
 enabled_metadata=1
+priority=999
 _EOF'
-            RHEL_EXTRA_PKGS="cfssl python3-pip"
             if rlIsRHEL 8 || rlIsCentOS 8; then
+                RHEL_EXTRA_PKGS="cfssl python3-pip"
                 RHEL_EXTRA_PIP_PKGS="packaging"
             else
-                RHEL_EXTRA_PKGS="$RHEL_EXTRA_PKGS python3-packaging"
+                RHEL_EXTRA_PKGS="cfssl python3-packaging python3-lark-parser"
```
sergio-correia, line 41: We removed support for `cfssl` a while ago, so this can go away as well

kkaarreell: I thought it is necessary for Python agent. If not, I will remove it.

sergio-correia: It's not. Here's the PR that removed support for it: https://github.com/keylime/keylime/pull/1012

PR #281, opened 2022-12-15
Use Sergio's copr repo only for RHEL8

kkaarreell:


sergio-correia (approved):
LGTM. Apprently, the only package we install from there is the IBM tpm emulator.

PR #284, opened 2022-12-20
Modify tests for keylime PR 1260

ansasaki:
This modifies the `upstream/run_keylime_tests` to include the `templates` directory used during the `test_restful.py` execution to generate the configuration files.

### DIFF
--- a/upstream/run_keylime_tests/test.sh
+++ b/upstream/run_keylime_tests/test.sh
@@ -36,7 +36,7 @@ rlJournalStart
             rlRun "ln -s ~/rpmbuild/BUILD/keylime* /var/tmp/keylime_sources"
         fi
         rlRun "TmpDir=\$( mktemp -d )"
-        rlRun "cp -r /var/tmp/keylime_sources/{test,test-data,tpm_cert_store,scripts} $TmpDir"
+        rlRun "cp -r /var/tmp/keylime_sources/{test,test-data,tpm_cert_store,scripts,templates} $TmpDir"
         pushd $TmpDir/test
         # if TPM emulator is present
         if limeTPMEmulated; then


ansasaki:
Currently it is not possible to build `tss-esapi` crate on Fedora Rawhide. The latest version of the crate is incompatible with the `tpm2-tss` package version 4.0 shipped on Fedora Rawhide, which is a pre-release version.

See: https://github.com/parallaxsecond/rust-tss-esapi/issues/382

ansasaki:
@Koncpa @kkaarreell Could you review this one-liner change? If you approve, I'll drop the commit that changes the tested branch to make it ready to merge.

Koncpa (approved):

ansasaki:
I dropped the commit that changes the tested branch.
Note that this makes the tests to fail as the `templates` directory was moved as part of the tested PR.

kkaarreell:
I think we should be copying only directories that exist

ansasaki:
> I think we should be copying only directories that exist. That way we won't unnecessarily diverge from RHEL9

The alternatives  I see is to not copy files at all in `upstream/run_keylime_tests` and run the tests directly from the sources directory (or the extreme opposite, copy everything).

TBH, I did not investigate why do we copy the files to a temporary dir (I guess the goal is to not pollute the source directory).

kkaarreell:
> TBH, I did not investigate why do we copy the files to a temporary dir (I guess the goal is to not pollute the source directory).

Yes. I think let's copy everything.. It seems to be 20 times slower but still under 200ms on my local system. :-)

kkaarreell:
Proposed in https://github.com/RedHat-SP-Security/keylime-tests/pull/285

PR #285, opened 2023-01-05
Copy all keylime_sources dir content

kkaarreell:
Let's copy all keylime_sources dir content so we don't have to touch this code whenever new directory is required.


ansasaki (approved):

kkaarreell:
These test failures on f36 and f37 are caused by tpm2-tss update that seemed to introduce a regression. Not related to this patch.

PR #288, opened 2023-01-10
Basic attestation scenario - agent,verifier, registrar in container

Koncpa:
Basic attestation scenario, where agent,verifier and registrar are in containers. Small changes in limeLib in name of function for working with containers and removing duplicated option from function.


container/functional/keylime_verifier_registrar_container-basic-attestation/test.sh
```diff
@@ -0,0 +1,123 @@
...
+        TAG_REGISTRAR="registrar_image"  
+        podman build -t=$TAG_REGISTRAR --file "${limeLibraryDir}/Dockerfile.registrar" .
+
+        #run verifier container
+        CONT_VERIFIER="verifier_container"
+        rlRun "podman run -d --name $CONT_VERIFIER --net $CONT_NETWORK_NAME --ip $IP_VERIFIER -p 8881:8881 --volume=/etc/keylime/:/etc/keylime/ localhost/$TAG_VERIFIER /usr/bin/keylime_verifier"
+        rlRun "limeWaitForVerifier"
+        #wait for generating of certs
+        sleep 5
+        rlRun "podman cp $CONT_VERIFIER:/var/lib/keylime/cv_ca/ ."
```
kkaarreell, line 49: We should discuss if it is better to copy certs like this or have that directory shared.

container/functional/keylime_verifier_registrar_container-basic-attestation/test.sh
```diff
@@ -0,0 +1,123 @@
...
+        TAG_VERIFIER="verifier_image"  
+        podman build -t=$TAG_VERIFIER --file "${limeLibraryDir}/Dockerfile.verifier" .   
+        
+        #build registrar container
+        TAG_REGISTRAR="registrar_image"  
+        podman build -t=$TAG_REGISTRAR --file "${limeLibraryDir}/Dockerfile.registrar" .
+
+        #run verifier container
+        CONT_VERIFIER="verifier_container"
+        rlRun "podman run -d --name $CONT_VERIFIER --net $CONT_NETWORK_NAME --ip $IP_VERIFIER -p 8881:8881 --volume=/etc/keylime/:/etc/keylime/ localhost/$TAG_VERIFIER /usr/bin/keylime_verifier"
```
kkaarreell, line 45: I am not sure if publishing port 8881 is right. Should we rather let services talk through the created network?

container/functional/keylime_verifier_registrar_container-basic-attestation/test.sh
```diff
@@ -0,0 +1,123 @@
...
+        rlRun "limeWaitForVerifier"
```
kkaarreell, line 46: This works only because we have published its port previously. Maybe we should introduce IP as an optional argument so that we would be waiting for an opened port on the given IP (and not always/only localhost).

container/functional/keylime_verifier_registrar_container-basic-attestation/test.sh
```diff
@@ -0,0 +1,123 @@
...
+        sleep 5
+        rlRun "podman cp $CONT_VERIFIER:/var/lib/keylime/cv_ca/ ."
+        
+        #tenant need certs
+        rlRun "cp -r cv_ca/ /var/lib/keylime/"
+
+        #run registrar container
+        CONT_REGISTRAR="registrar_container"
+        rlRun "podman run -d --name $CONT_REGISTRAR --net $CONT_NETWORK_NAME --ip $IP_REGISTRAR -p 8891:8891 -p 8890:8890 --volume=/etc/keylime/:/etc/keylime/ --volume=$PWD/cv_ca:/var/lib/keylime/cv_ca:rw localhost/$TAG_REGISTRAR /usr/bin/keylime_registrar"
+        rlRun "limeWaitForRegistrar"
```
kkaarreell, line 57: Same thing regarding publishing ports as for the verifier above.

container/functional/keylime_verifier_registrar_container-basic-attestation/test.sh
```diff
@@ -0,0 +1,123 @@
...
+        rlRun "limeUpdateConf tenant registrar_ip $IP_REGISTRAR"
+
+        #setup of agent
+        TAG_AGENT="agent_image"
+        CONT_AGENT="agent_container"
+        rlRun "cp cv_ca/cacert.crt ."
+        rlRun "limeconPrepareAgentImage ${TAG_AGENT}"
+        rlRun "limeUpdateConf agent registrar_ip '\"$IP_REGISTRAR\"'"
+        rlRun "limeconPrepareAgentConfdir $AGENT_ID $IP_AGENT confdir_$CONT_AGENT"
+        rlRun "chmod o+rw /dev/tpmrm0"
```
kkaarreell, line 71: Not sure if it matters for a test purposes. Maybe we should be running agent container under a user that can access it. Definitely this is not a good practice in general.

kkaarreell (commented):
I am happy this works. However I am not sure about some constructions, maybe we should think more about our use cases for these containers and adjust our approach according to that. I will leave this open for now, let's discuss it in person.

container/functional/keylime_verifier_registrar_container-basic-attestation/main.fmf
```diff
@@ -0,0 +1,32 @@
...
+tag:
+  - container
+component:
+  - keylime
+test: ./test.sh
+framework: beakerlib
+require:
+  - yum
+  - podman
+  - nmap
```
kkaarreell, line 23: We are importing `test-helpers` library directly and therefore `require` of the library itself won't have an effect.. And therefore you are adding it to a test.
I am thinking, what about installing all library requires during the library import so we won't have to add `nmap` to every test using it?

Library/test-helpers/lib.sh
```diff
@@ -843,6 +843,88 @@ limeStopTPMEmulator() {
...
+
+    if [ -z "${IP}" ]; then
+        if ! rlWaitForSocket $PORT -d 0.5 -t ${limeTIMEOUT}; then
+            cat $LOGFILE
+            return 1
+        else
+            return 0
+        fi
+    elif [ -n "${IP}" ]; then
+        rlWaitForCmd "limeCheckRemotePort ${PORT} ${IP}" -m 3 -t 3 -d 1
```
kkaarreell, line 925: use `-t ${limeTIMEOUT}` instead of `-m 3 -t 3 -d 1` so it is in par with the command above. Sane default for `d` is probably 1s (since it is network connection).

Library/test-helpers/lib.sh
```diff
@@ -843,6 +843,88 @@ limeStopTPMEmulator() {
...
+    local IP=$3
+
+    if [ -z "${IP}" ]; then
+        if ! rlWaitForSocket $PORT -d 0.5 -t ${limeTIMEOUT}; then
+            cat $LOGFILE
+            return 1
+        else
+            return 0
+        fi
+    elif [ -n "${IP}" ]; then
```
kkaarreell, line 924: this condition is not necessary since it cannot be otherwise.

Library/test-helpers/lib.sh
kkaarreell: Also, maybe close port and IP in quotes to avoid unwanted expansion by shell, i.e.
```
rlWaitForCmd "limeCheckRemotePort '${PORT}' '${IP}'" -d 1 -t ${limeTIMEOUT}
```

container/functional/keylime_verifier_registrar_container-basic-attestation/test.sh
kkaarreell: Please move this `chmod` to `limeconRunAgent` so we have it just on a single place and can fix it there in the future.

kkaarreell: Also, I am thinking whether we should update `limeBackupConfig` to save current permissions and restore them later in `limeRestoreConfig`

kkaarreell (changes requested):
just few nitpicks.

Library/test-helpers/lib.sh
```diff
@@ -1950,11 +2198,15 @@ Returns 0.
...
 
-limeconStopAgent() {
+limeconStop() {
 
         local NAMES=$1
 
         podman ps -a --format "{{.Names}}" | grep -e "^${NAMES}\$" | xargs podman stop | xargs podman rm
+
+        if [[ "agent_container" = *"$NAMES"* ]];then
+            chmod o-rw /dev/tpmrm0
```
kkaarreell, line 2208: We cannot do that here because there could be other agents still running (with multiple agents scenario). Moreover `agent_container` is just a specific name used in this particular test, not a generic one. Either do this in a setup&cleanup phases in a container test itself (my choice for now) or in the `limeBackupConfig` and `limeRestoreConfig` or we could eventually introduce new `limeconBackupConfig` and `limeconBackupConfig` or something similar.

Library/test-helpers/lib.sh
```diff
@@ -2032,6 +2284,9 @@ limeLibraryLoaded() {
         rlLogError "Failed loading library keylime/test-helpers."
         return 1
     fi
+
+    echo -e "\nInstalled requiered packages for helping library"
```
kkaarreell, line 2288: typo in "required"

Library/test-helpers/lib.sh
```diff
@@ -2032,6 +2284,9 @@ limeLibraryLoaded() {
...
+    dnf install -y nmap
```
kkaarreell, line 2289: can you add also other packages used in a library (tpm2-tools, openssl, beakerlib, maybe podman???) and install them only when they are not installed?
`rpm -q $PACKAGES || yum -y install $PACKAGES`

kkaarreell (changes requested):

Koncpa:
Now should be all nitpicks resolved, so I think that we can merge it. @kkaarreell

Library/test-helpers/lib.sh
```diff
@@ -2032,6 +2280,11 @@ limeLibraryLoaded() {
         rlLogError "Failed loading library keylime/test-helpers."
         return 1
     fi
+
+    local PACKAGES="tpm2-tools openssl beakerlib podman"
+
+    echo -e "\nInstalled required packages for helping library if missing."
```
kkaarreell, line 2286: weird wording, pls use: Install packages required by the library when missing.

Library/test-helpers/lib.sh
```diff
@@ -2032,6 +2280,11 @@ limeLibraryLoaded() {
         rlLogError "Failed loading library keylime/test-helpers."
         return 1
     fi
+
+    local PACKAGES="tpm2-tools openssl beakerlib podman"
```
kkaarreell, line 2284: you forgot that nmap package

Library/test-helpers/lib.sh
```diff
@@ -1884,7 +2045,93 @@ limeconRunAgent() {
     local AGENT_FILE=$5
     local TESTDIR=$6
 
-    podman run -d --name $NAME --net $NETWORK --ip ${IP} --privileged --volume=${AGENT_FILE}:/etc/keylime/ --volume=/sys/kernel/security/:/sys/kernel/security/:ro --volume=${TESTDIR}:${TESTDIR}:rw --ip ${IP} --device=/dev/tpmrm0  localhost/$TAG /usr/bin/keylime_agent
+    #mandatory for access containers to tpm
+    chmod o+rw /dev/tpmrm0
```
kkaarreell, line 2049: Pls move this to the test itself so both `o+rw` and restore `o-rw` are at the same place, otherwise it would be confusing for a reader.

kkaarreell:
> Now should be all nitpicks resolved, so I think that we can merge it. @kkaarreell

Last 3 things, I promise. :-)

Library/test-helpers/lib.sh
Koncpa: Sure, that sounds better.

Library/test-helpers/lib.sh
Koncpa: Oh I forgot about that package.

Library/test-helpers/lib.sh
Koncpa: Yeah, that could be more readable for others.

kkaarreell:
@Koncpa Regarding the update due to IMA policy overhaul, please do this change in a separate commit so we can cherry-pick container updates (without overhaul) into rhel-9-main and fedora-rawhide branches.

kkaarreell (approved):

Koncpa:
Manual run PASS, merging.

PR #305, opened 2023-01-26
Do not update /sys/kernel/security/ima/policy in setup task

kkaarreell:


kkaarreell:
This change does not have desired effect so I am closing it.

PR #307, opened 2023-02-09
Disable agent uuid assignment test due to bz#2165055

Koncpa:
Remove for all distros except f36 and f37
bz#2165055

### DIFF
--- a/functional/agent_UUID_assignment_options/test.sh
+++ b/functional/agent_UUID_assignment_options/test.sh
@@ -49,15 +49,18 @@ rlJournalStart
         rlRun "limeStopAgent"
     rlPhaseEnd
 
+    # disable agent_UUID_assignment_options test regarding to bz#2165055
+    if false; then
     rlPhaseStartTest "Generate UUID"
         AGENT_ID="generate"
         rlRun "limeUpdateConf agent  uuid '\"$AGENT_ID\"'"
         rlRun "limeStartAgent"
         sleep 3
-        GENERATE_UUID=$(systemctl status keylime_agent -all | tail -2 | head -1 | grep -oP '(?<=Agent )[^ ]*')
+        GENERATE_UUID=$(journalctl _SYSTEMD_INVOCATION_ID=$(systemctl show -p InvocationID --value keylime_agent.service) | grep -oP '(?<=Generated a new UUID: ).*')
         rlRun "limeWaitForAgentRegistration ${GENERATE_UUID}"
         rlRun "limeStopAgent"
     rlPhaseEnd
+    fi
 
     rlPhaseStartTest "Use ENV variable to set agent UUID"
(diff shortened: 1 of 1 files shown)


functional/agent_UUID_assignment_options/test.sh
```diff
@@ -49,15 +49,18 @@ rlJournalStart
...
 
+    # disable agent_UUID_assignment_options test regarding to bz#2165055
+    if [ -n "$UUID_PHASE" ]; then
     rlPhaseStartTest "Generate UUID"
         AGENT_ID="generate"
         rlRun "limeUpdateConf agent  uuid '\"$AGENT_ID\"'"
         rlRun "limeStartAgent"
         sleep 3
-        GENERATE_UUID=$(systemctl status keylime_agent -all | tail -2 | head -1 | grep -oP '(?<=Agent )[^ ]*')
+        GENERATE_UUID=$(journalctl _SYSTEMD_INVOCATION_ID=`systemctl show -p InvocationID --value keylime_agent.service` | grep -oP '(?<=Generated a new UUID: ).*')
```
kkaarreell, line 59: please use
```
$( systemctl show -p InvocationID --value keylime_agent.service) 
```
instead of
```
`systemctl show -p InvocationID --value keylime_agent.service`
```
as it is better readable when commands are nested.

functional/agent_UUID_assignment_options/test.sh
```diff
@@ -49,15 +49,18 @@ rlJournalStart
         rlRun "limeStopAgent"
     rlPhaseEnd
 
+    # disable agent_UUID_assignment_options test regarding to bz#2165055
+    if [ -n "$UUID_PHASE" ]; then
```
kkaarreell, line 53: We are not working with this variable, better use simply `if false; then` to make it obvious the phase is intentionally disabled.

kkaarreell (changes requested):

functional/agent_UUID_assignment_options/test.sh
Koncpa: Yeah, thanks that's more handy. I added variable due to manual testing, if we setting variable in plan we could run this subtest. But nevermind, we can simply remove the conditional.

kkaarreell: Right, I got its purpose. But the thing is, in order to understand what `UUID_PHASE` variable defined in a test plan really means, one would have to remember it relates to this specific test end enables/disables a subtest. IMHO we would have give this variable more descriptive name and even that, I am not sure it is really worth it.

kkaarreell (approved):

kkaarreell:
Test failure is expected, the tenant related test fails due to https://github.com/keylime/keylime/issues/1306 and not related to this PR.
Let's merge it.

PR #312, opened 2023-02-24
Implement tests for DSSE signing

mbestavros:
[Keylime PR 1314](https://github.com/keylime/keylime/pull/1314) changes how runtime policies are signed. This PR updates policy creation to generate DSSE-signed policies and test them.

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -1374,8 +1374,11 @@ limeCreateTestPolicy() {
     cat $__INTERNAL_limeBaseExcludeList >> excludelist.txt
     echo -e "${EXCLUDE}" >> excludelist.txt
 
-    # create policy.json...
-    rlRun "limeCopyKeylimeFile --install cmd/convert_runtime_policy.py && python3 convert_runtime_policy.py -a allowlist.txt -e excludelist.txt -o policy.json"
+    # create policy.json and create signed policies and keys
+    keylime_convert_runtime_policy -a allowlist.txt -e excludelist.txt -o policy.json && \
+    keylime_sign_runtime_policy -r policy.json -p dsse-ecdsa-privkey.key -b ecdsa -o policy-dsse-ecdsa.json && \
+    keylime_sign_runtime_policy -r policy.json -p dsse-x509-privkey.key -b x509 -o policy-dsse-x509.json && \
+    openssl ec -in dsse-ecdsa-privkey.key -pubout -out dsse-ecdsa-pubkey.pub
 
 }
 
@@ -2192,7 +2195,7 @@ true <<'=cut'
 
 =head2 limeconStop
 
-Stop container, delete container and set default permission 
+Stop container, delete container and set default permission
 for agent container if stopping agent container.
 
     limeconStop NAMES
--- a/functional/tenant-runtime-policy-sanity/test.sh
+++ b/functional/tenant-runtime-policy-sanity/test.sh
@@ -38,37 +38,14 @@ rlJournalStart
         rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
         # create allowlist and excludelist
         rlRun "limeCreateTestPolicy /etc/hostname"
+        rlRun "sed 's/0/1/' policy-dsse-ecdsa.json > policy-dsse-ecdsa-bad.json"
         CHECKSUM=$( sha256sum policy.json | cut -d ' ' -f 1 )
+        CHECKSUM_DSSE_ECDSA=$( sha256sum policy-dsse-ecdsa.json | cut -d ' ' -f 1 )
+        CHECKSUM_DSSE_X509=$( sha256sum policy-dsse-x509.json | cut -d ' ' -f 1 )
+        CHECKSUM_DSSE_ECDSA_BAD=$( sha256sum policy-dsse-ecdsa-bad.json | cut -d ' ' -f 1 )
         # start simple http server to serve files
         rlRun "python3 -m http.server 8000 &"
         HTTP_PID=$!
-        # generate GPG key
-        rlRun "gpgconf --kill gpg-agent"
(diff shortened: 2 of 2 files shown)


Koncpa:
Looks like that tenant need also use --runtime_policy-checksum, when is set --runtime_policy-url option.

mbestavros:
@Koncpa Thanks, that worked!

The `rawhide` failure seems bigger than my changes...

Koncpa:
@mbestavros  LGTM!

Failures in Fedora Rawhide are expected, due to:
https://github.com/tpm2-software/tpm2-tools/issues/3210
https://bugzilla.redhat.com/show_bug.cgi?id=2171376

Library/test-helpers/lib.sh
```diff
@@ -1376,6 +1376,10 @@ limeCreateTestPolicy() {
 
     # create policy.json...
     rlRun "limeCopyKeylimeFile --install cmd/convert_runtime_policy.py && python3 convert_runtime_policy.py -a allowlist.txt -e excludelist.txt -o policy.json"
+    # create signed policies and keys
+    rlRun "limeCopyKeylimeFile --install cmd/sign_runtime_policy.py && python3 sign_runtime_policy.py -r policy.json -p dsse-ecdsa-privkey.key -b ecdsa -o policy-dsse-ecdsa.json"
```
kkaarreell, line 1380: @mbestavros Where is the dsse-ecdsa-privkey.key coming from?

Library/test-helpers/lib.sh
```diff
@@ -1376,6 +1376,10 @@ limeCreateTestPolicy() {
...
+    rlRun "limeCopyKeylimeFile --install cmd/sign_runtime_policy.py && python3 sign_runtime_policy.py -r policy.json -p dsse-x509-privkey.key -b x509 -o policy-dsse-x509.json"
```
kkaarreell, line 1381: Would it be possible to deliver this script on a system using the common keylime installation script? This way we won't have to copy it here and moreover it would be included in code coverage measurement.

Library/test-helpers/lib.sh
mbestavros: @kkaarreell It's generated as part of the signing script. A user can pass an ECDSA private key, or the tool can create one if not specified.

Library/test-helpers/lib.sh
mbestavros: @kkaarreell How would one do that? Do any scripts in [`setup.cfg`](https://github.com/keylime/keylime/blob/master/setup.cfg) work?

kkaarreell: I believe so. We are installing keylime through `python3 setup.py install`.

kkaarreell: I mean, I am assuming users would be using this script too and therefore it makes sense to install it but I may be wrong.

Library/test-helpers/lib.sh
kkaarreell: In fact I can see that in tests we copy scripts here so maybe you can simply run it from this path
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/setup/install_upstream_keylime/test.sh#L91

Library/test-helpers/lib.sh
kkaarreell: @mbestavros I am sorry, I did more mess with my comments. I somehow confused `cmd` with `scripts`. We are installing `scripts` to `/usr/share/keylime/scripts` but this does not apply to `cmd` which we expect will be installed to `/usr/bin`. We can either leave it as it is now but if you expect that users would be using this script (I would) then I would propose to install it on a system through the setup script (and maybe consult other devs regarding other installation options). Also, I would consider renaming it (add it `keylime_` prefix and remove `.py` suffix) so the naming is consistent with other keylime scripts.

mbestavros: @kkaarreell I've just tested manually building using `python setup.py install` and the signing script is accessible via `keylime_sign_runtime_policy`. So I can edit my tests to use that command.

mbestavros: @kkaarreell I've updated `limeCreateTestPolicy` to use `keylime_sign_runtime_policy` in both new cases. I've also updated the conversion to use `keylime_convert_runtime_policy` using the same approach.

kkaarreell (approved):
Thank you very much for these test updates.
Now we should wait before the keylime upstream PR is ready for merge.
After removing the redirect in `main.fmf` we can merge it.

kkaarreell:
Hi @mbestavros , I had to do some updates in keylime-tests due to the Python agent removal. Could you please rebase your PR to ensure it is tested in CI properly? Thank you.

Library/test-helpers/lib.sh
```diff
@@ -1375,7 +1375,11 @@ limeCreateTestPolicy() {
     echo -e "${EXCLUDE}" >> excludelist.txt
 
     # create policy.json...
-    rlRun "limeCopyKeylimeFile --install cmd/convert_runtime_policy.py && python3 convert_runtime_policy.py -a allowlist.txt -e excludelist.txt -o policy.json"
+    rlRun "keylime_convert_runtime_policy -a allowlist.txt -e excludelist.txt -o policy.json"
```
kkaarreell, line 1378: Could you please remove `rlRun` from this and following calls and possibly also connect those individual commands together and with the previous code through `&&`? We are calling this library function in tests from `rlRun` so better there are not two nested `rlRun`s used.

kkaarreell:
Please rebase.

mbestavros:
@kkaarreell Whoops, my mistake. I've also dropped the nested `rlRun` calls in my latest push.

kkaarreell:
Thank you @mbestavros , I think we are safe to proceed with dropping the redirect commit. I will merge it right away (since tests would be failing anyway).

PR #318, opened 2023-03-05
Changes required for the agent removal

THS-on:


THS-on:
@kkaarreell I dropped the removal of the unused python payloads for now, so it should be fine to merge before the agent removal is done.

kkaarreell:
```
ModuleNotFoundError: No module named 'toml'
```
All the impacted tests needs to be updated with the respective package in `main.fmf`. I guess it is `python3-toml`.

THS-on:
@kkaarreell yes I originally had that, but forgot to re-add that. Should now be fixed

kkaarreell:
I think that the test failure is actually a bug in a test itself. `tpm2-abrmd` service start should be before `tpm2_nvread` IMHO.

kkaarreell:
> I think that the test failure is actually a bug in a test itself. `tpm2-abrmd` service start should be before `tpm2_nvread` IMHO.
https://github.com/RedHat-SP-Security/keylime-tests/pull/321
@THS-on Would you mind taking a look?

kkaarreell:
@THS-on We need to update every single test using `toml` with that new dependency. I can eventually do that in another PR later but if we won't that it won't be possible to run such a test as a standalone one.

THS-on:
@kkaarreell i think we can merge this

PR #319, opened 2023-03-06
Simplify rust keylime install task

kkaarreell:
Due to upcoming Python keylime agent removal we should follow the common Rust agent installation path so that executables and systemd unit files land on a system properly.

### DIFF
--- a/setup/install_upstream_rust_keylime/test.sh
+++ b/setup/install_upstream_rust_keylime/test.sh
@@ -2,8 +2,6 @@
 # vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
 . /usr/share/beakerlib/beakerlib.sh || exit 1
 
-# define RUST_IMA_EMULATOR variable to install also rust IMA emulator
-
 [ -n "${RUST_KEYLIME_UPSTREAM_URL}" ] || RUST_KEYLIME_UPSTREAM_URL="https://github.com/keylime/rust-keylime.git"
 [ -n "${RUST_KEYLIME_UPSTREAM_BRANCH}" ] || RUST_KEYLIME_UPSTREAM_BRANCH="master"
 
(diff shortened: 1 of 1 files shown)


setup/install_upstream_rust_keylime/test.sh
```diff
@@ -67,38 +55,34 @@ _EOF'
         rlRun "cp tests/actions/shim.py /usr/libexec/keylime"
```
kkaarreell, line 55: Not sure if the shim related part is still necessary. The upstream issue is still open.

setup/install_upstream_rust_keylime/test.sh
```diff
@@ -40,19 +38,9 @@ rlJournalStart
         fi
         #build
         rlRun "cargo build"
```
kkaarreell, line 40: Should we just simply run `make` instead?

Koncpa: I'm not sure I just follow previous setup of test.. but I would say that will be better to use `cargo` as specialized tool for rust.

setup/install_upstream_rust_keylime/test.sh
Koncpa: Do you have link to this upstream issue?

setup/install_upstream_rust_keylime/test.sh
kkaarreell: Makefile also users cargo. I am thinking that by using `make` we would also test the Makefile itself to some extent. I will update it.

setup/install_upstream_rust_keylime/test.sh
kkaarreell: https://github.com/keylime/rust-keylime/issues/325 It is mentioned in a comment above.

setup/install_upstream_rust_keylime/test.sh
Koncpa: Sure, I know that `make` can be used also for rust project I just want to pointed out that `cargo` has a much deeper understanding of Rust.

kkaarreell: What I mean is that `cargo` is used in the Makefile.
https://github.com/keylime/rust-keylime/blob/master/GNUmakefile#L26

PR #325, opened 2023-03-08
Add simple sanity check for durable attestation

maugustosilva:
This test was pieced together with examples from
measured-boot-swtpm-sanity and basic-attestation-with-ima-signatures. The main idea here is to add an agent with both measured boot and runtime policies, and make sure offline attestation works on it.

When `keylime_attest` is run, all that matters is the exit code (0 will indicate that all stored attestation artifacts were properly attested)


maugustosilva:
@kkaarreell it looks like `tsseventextend` is no longer part of FC37. Do you have any suggestions on how to implement the test with an agent that supplies a measured boot log?

kkaarreell:
> @kkaarreell it looks like `tsseventextend` is no longer part of FC37. Do you have any suggestions on how to implement the test with an agent that supplies a measured boot log?

Hi @maugustosilva , it was F38 where the package was missing. Anyway, few weeks back I have approached RH packages with this problem and they have replied that they have a new Fedora maintainer for this package. Looking into Koji it seems that some builds are already landing and I can install this package already on my test system.
https://koji.fedoraproject.org/koji/packageinfo?packageID=23163
```
Dependencies resolved.
================================================================================
 Package     Architecture  Version                 Repository              Size
================================================================================
Installing:
 tss2        x86_64        1:1.6.0-6.fc38          updates-testing        608 k

Transaction Summary
================================================================================
Install  1 Package

Total download size: 608 k
Installed size: 2.6 M
Is this ok [y/N]:                
```

kkaarreell:
Btw, we have disabled tests for Rawhide because of the [tpm2-tools issue](https://github.com/tpm2-software/tpm2-tools/issues/3210) so F37 will be the latest tested Fedora for some time.

functional/durable-attestion-sanity-on-localhost/test.sh
```diff
@@ -0,0 +1,99 @@
...
+
+    rlPhaseStartSetup "Do the keylime setup"
+        rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
+        rlAssertRpm keylime
+        
+        # clean previous durable attestion artifacts stored on filesystem
+        rlRun "rm -rf /var/lib/keylime/da/CI_*"
+
+        # copy fake binary_bios_measurements into an alternative path (used to create refstate)
+        rlRun "cp binary_bios_measurements /var/tmp"
```
kkaarreell, line 18: I believe you are not changing a working directory so this copy is probably not necessary.

functional/durable-attestion-sanity-on-localhost/main.fmf
```diff
@@ -0,0 +1,22 @@
...
+contact: Marcio Silva <marcio.a.silva@ibm.com>
+component:
+  - keylime
+test: ./test.sh
+framework: beakerlib
+tag:
+  - CI-Tier-1
+require:
+  - yum
+  - tpm2-tools
```
kkaarreell, line 18: add ` - tss2` to get the tsseventextend package installed.

functional/durable-attestion-sanity-on-localhost/test.sh
kkaarreell: Ignore my comment, it is necessary.

functional/durable-attestion-sanity-on-localhost/main.fmf
```diff
@@ -0,0 +1,22 @@
...
+framework: beakerlib
+tag:
+  - CI-Tier-1
+require:
+  - yum
+  - tpm2-tools 
+recommend:
+  - keylime
+duration: 5m
+extra-nitrate: TC#0614117
```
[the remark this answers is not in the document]
kkaarreell: Also, we should add additional test metadata. Please copy here whole `adjust` block from
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/functional/measured-boot-swtpm-sanity/main.fmf#L27

kkaarreell (changes requested):
Thank you for the test. Please, incorporate suggested changes, they are necessary so we can properly execute the test from our test plans.

functional/durable-attestion-sanity-on-localhost/main.fmf
maugustosilva: fixed

functional/durable-attestion-sanity-on-localhost/main.fmf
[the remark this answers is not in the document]
maugustosilva: done

kkaarreell (approved):
LGTM, thank you.

kkaarreell:
@maugustosilva What would be the right procedure to test also durable attestation failure? Should we simply let the agent fail attestation (e.g. by running a modified script) and once it fails the check on the verifier it should also fail durable attestation?

maugustosilva:
> @maugustosilva What would be the right procedure to test also durable attestation failure? Should we simply let the agent fail attestation (e.g. by running a modified script) and once it fails the check on the verifier it should also fail durable attestation?

Ah, very good question. We could add an agent which will fail (I have noticed this is done for a few of the tests already), then execute `keylime_attest` and expect for a non-zero exit code.

kkaarreell:
Could you possibly extend this PR with such a scenario? We are failing agent e.g. here
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/functional/basic-attestation-on-localhost/test.sh#L92
I think you can exclude lines 99-107.

Koncpa (approved):

kkaarreell:
> Could you possibly extend this PR with such a scenario? We are failing agent e.g. here https://github.com/RedHat-SP-Security/keylime-tests/blob/main/functional/basic-attestation-on-localhost/test.sh#L92 I think you can exclude lines 99-107.

I will create a separate PR with that.

maugustosilva:
I have been pulled into other tasks for a few days, came back here this very morning to do it, and saw that it is already done and merged. Thanks a lot @kkaarreell !

kkaarreell:
Hi @maugustosilva, I was also thinking, It shouldn't matter if keylime services are running or not, right? Would it make sense to rerun check again with services stopped?

maugustosilva:
Very good question: `keylime_attest` specifically and purposefully does NOT require any Keylime services (DB, Verifier or Registrar)

PR #335, opened 2023-03-10
Measure code coverage for keylime-agent-rust SRPM

Koncpa:


kkaarreell:
This PR (and commit) should be named: Measure code coverage for keylime-agent-rust SRPM.
It is not related to patch coverage.

PR #338, opened 2023-03-14
Improve patch coverage precision

kkaarreell:
Should improve detection of a code change inside a patch and avoid counting `}` as not executed lines.


kkaarreell:
@Koncpa  It is not really easy to test this. I have tested it locally manually feeding test with `PACKIT_` variables to simulate real GitHub PR environment. As far as I can tell it is working fine.

kkaarreell:
We may need to backport this to rhel-9-main and rhel-9.1.0 branches.

PR #340, opened 2023-03-14
Run tier tests on Testing Farm

ansasaki:
Add a reusable workflow that schedule tests on testing farm using the CLI. If the specified plan is not found, the test is ignored and passes.

The tier tests are executed sequentially using the reusable workflow. The tests from the tier1.yaml plan are executed on a single distro, currently Fedora 37.

If the 'tier' input is set as 'tier2', the tests from the tier2.yaml plan are executed in multiple distros, currently Fedora 37 and 38.

The tests are triggered when a pull request is opened or reopened.


ansasaki:
To see what happens, I merged this PR in my own repo: https://github.com/ansasaki/keylime-tests/

Check what happens when a PR is opened: https://github.com/ansasaki/keylime-tests/pull/1

Then, after the PR is registered successfully, it will trigger the internal workflow: https://github.com/ansasaki/keylime-tests/actions/runs/5142302701

Currently, the ["check" API](https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?apiVersion=2022-11-28) is not implemented to link the second workflow with the PR that originated the run.

PR #345, opened 2023-03-16
Add 2 test cases expecting command rejection due to a malformed runtime policy

stefanberger:


kkaarreell (approved):
Thank you for these additions, I will merge it once tests on F36 pass. It is failing due to an infrastructure issue.

stefanberger:
> Thank you for these additions, I will merge it once tests on F36 pass. It is failing due to an infrastructure issue.

The tenant tool's {add/update/delete}runtimepolicy commands currently do not handle errors returned from the server correctly and once they do that the 3 commented out lines can be activated but other test cases will have to be adjusted as well.

kkaarreell:
Is there some issue tracking it?

PR #348, opened 2023-03-17
Fix broken rust coverage measurement

kkaarreell:
This fixes broken coverage measurement due to changed filepaths and absent Rust flags. Also, always use Rustup to avoid duplicite Rust installation.

### DIFF
--- a/setup/install_upstream_rust_keylime/test.sh
+++ b/setup/install_upstream_rust_keylime/test.sh
@@ -10,7 +10,7 @@ rlJournalStart
     rlPhaseStartSetup "Build and install rust-keylime bits"
         rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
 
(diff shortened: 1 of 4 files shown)


setup/install_upstream_rust_keylime/test.sh
```diff
@@ -10,7 +10,7 @@ rlJournalStart
     rlPhaseStartSetup "Build and install rust-keylime bits"
         rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
 
-        if [ ${RPM_AGENT_COVERAGE} == "1" ] && rpm -qa | grep keylime-agent-rust; then
+        if [ "${RPM_AGENT_COVERAGE}" == "1" ] && rpm -qa | grep keylime-agent-rust; then
```
stefanberger, line 13: I think it should be '=' rather than '=='

kkaarreell:
@ansasaki please review.

PR #359, opened 2023-03-31
Run container with CAP's to allow ssh access

Koncpa:
Verifier and registrar containers deny ssh access due to insufficient permission. It can be allowed when run container with capabilities.


kkaarreell (approved):

kkaarreell:
@Koncpa did you tested it already? Are these updates sufficient to make sshd to work?

Koncpa:
Yeah, it was manually tested and I can prove that this solution is sufficient.

PR #360, opened 2023-04-04
Optimization container functions in helper lib

Koncpa:
limeconPrepareImage function replace prepare image functions for specific type of container. Now is
mandatory to specify also docker file for each
call of function. limeconRun is new core function
for start of container. Is called by specific limeconRun function which add setup for specific type of container (agent, registrar, etc. )
Apply new approach to current tests and remove function for run systemd container.

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -1877,24 +1877,28 @@ limeconDeleteNetwork() {
 
     local NAME=$1
 
-    podman network rm $NAME
+    podman network rm -f $NAME
 }
 
 true <<'=cut'
 =pod
 
-=head2 limeconPrepareAgentImage
+=head2 limeconPrepareImage
 
-Prepare podman image with keylime agent. Install keylime agent and
-copy CA cert from the host into image and set needed permissions.
+Prepare podman image. Specify docker file and name tag for building images.
+Also set up ssh access for each container.
 
-    limeconPrepareAgentImage [-f|--file DOCKER_FILE] TAG
+    limeconPrepareImage [ --cacert CERT_FILE] DOCKER_FILE TAG
 
 =over
 
-=item -f, --file DOCKER_FILE
+=item --cacert CERT_FILE
 
-Parameter specify use of non-default docker file.
+Parameter specify path for cacert file.
+
+=item DOCKER_FILE
+
+Parameter specify use of docker file.
 
 =item TAG
 
@@ -1906,19 +1910,24 @@ Returns 0.
 
 =cut
 
-limeconPrepareAgentImage() {
+limeconPrepareImage() {
 
-    if [ "$1" == "-f" -o "$1" == "--file" ]; then
-        local DOCKER_FILE=$2
+
+    if [ "$1" == "--cacert" ]; then
+        local CERT_FILE=$2
+        cp $CERT_FILE .
         shift 2
-    else
-        echo "Using default docker file: ${limeLibraryDir}/Dockerfile.agent"
-        DOCKER_FILE="${limeLibraryDir}/Dockerfile.agent"
     fi
 
-    local TAG=$1
+    local DOCKER_FILE=$1
+    local TAG=$2
 
-    cp /var/lib/keylime/cv_ca/cacert.crt .    
+    if [ -z "${DOCKER_FILE}" ] || [ -z "${TAG}" ]; then
+        echo "Docker file or build tag was not specified!"
+        return 1
+    fi
+
+    #set up for ssh access
     ls /root/.ssh/id_*.pub || ssh-keygen -t rsa -N "" -f /root/.ssh/id_rsa
     cp /root/.ssh/id_*.pub .
     cp ${limeLibraryDir}/lime_con_start.sh .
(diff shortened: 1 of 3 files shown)


Library/test-helpers/lib.sh
```diff
@@ -1883,20 +1883,20 @@ limeconDeleteNetwork() {
...
-=head2 limeconPrepareAgentImage
+=head2 limeconPrepareImage
 
-Prepare podman image with keylime agent. Install keylime agent and
-copy CA cert from the host into image and set needed permissions.
+Prepare podman image. Specify docker file and name tag for building images.
+Also set up ssh access for each container.
 
-    limeconPrepareAgentImage [-f|--file DOCKER_FILE] TAG
+    limeconPrepareImage [-f|--file DOCKER_FILE] [-t|--tag TAG]
```
kkaarreell, line 1891: remove these `-f` and `-t` switches, we can work directly with positional arguments now when using them is mandatory. Your code below doesn't allow switching the options order anyway.

Library/test-helpers/lib.sh
```diff
@@ -1906,62 +1906,21 @@ Returns 0.
 
 =cut
 
-limeconPrepareAgentImage() {
+limeconPrepareImage() {
 
     if [ "$1" == "-f" -o "$1" == "--file" ]; then
```
kkaarreell, line 1911: Add check ensuring that exactly two arguments (dockerfile and tag) are given, fail otherwise. Then use `DOCKERFILE=$1` and `TAG=$2`

Library/test-helpers/lib.sh
```diff
@@ -1906,62 +1906,21 @@ Returns 0.
...
-        shift 2
-    else
-        echo "Using default docker file: ${limeLibraryDir}/Dockerfile.registrar"
-        DOCKER_FILE="${limeLibraryDir}/Dockerfile.registrar"
+    if [ "$3" == "-t" -o "$3" == "--tag" ]; then
+        local TAG=$4
     fi
 
-    local TAG=$1
+    if [[ "$DOCKER_FILE" =~ .*"Dockerfile.agent".* ]]; then
```
kkaarreell, line 1919: I do not like this hardcoded match.. I was thinking, we could add optiong like `--cacert FILE` which would ensure that a cert from the given path is copied into the keylime container.

Library/test-helpers/lib.sh
```diff
@@ -1971,86 +1930,52 @@ limeconPrepareRegistrarImage() {
...
+Name of used podman network.
 
-=over
+=item CMD_RUN
 
-=item -f, --file DOCKER_FILE
+Specify running command at start of container.
 
-Parameter specify use of non-default docker file.
+=item CMD_SETUP
```
kkaarreell, line 1959: Could we rather use something like `EXTRA_PODMAN_ARGS`? The content is not an actual command, just arguments.

Library/test-helpers/lib.sh
```diff
@@ -1971,86 +1930,52 @@ limeconPrepareRegistrarImage() {
...
-        DOCKER_FILE="${limeLibraryDir}/Dockerfile.systemd"
-    fi
+limeconRun() {
 
-    local TAG=$1
+    local NAME=$1
+    local TAG=$2
+    local IP=$3
+    local NETWORK=$4
+    local CMD_RUN=$5
```
kkaarreell, line 1975: I think now we could call it `COMMAND`.

kkaarreell (changes requested):

Library/test-helpers/lib.sh
```diff
@@ -2142,8 +2074,7 @@ limeconRunRegistrar() {
     local IP=$3
     local NETWORK=$4
 
-
-    podman run -d --name $NAME --net $NETWORK --ip $IP --cap-add CAP_AUDIT_WRITE --cap-add CAP_SYS_CHROOT --volume=/etc/keylime/:/etc/keylime/ --volume=$PWD/cv_ca:/var/lib/keylime/cv_ca:z localhost/$TAG /usr/local/bin/lime_con_start /usr/bin/keylime_registrar
+    limeconRun $NAME $TAG $IP $NETWORK "/usr/local/bin/lime_con_start /usr/bin/keylime_registrar" "--volume=/etc/keylime/:/etc/keylime/ --volume=$PWD/cv_ca:/var/lib/keylime/cv_ca:z"
```
kkaarreell, line 2077: I just realized, sometimes keylime scripts can be in `/usr/local/bin`, we should probably address that somehow.
We could e.g. pass only the script name (`keylime_registrar`) and do the path detection in the `lime_con_start` script (a.k.a. `SCRIPT_PATH=$( which $SCRIPT )`). Maybe it is not necessary to fix it right away but you will probably hit this problem when working on the kontainer with upstream keylime.

kkaarreell (approved):
I think this is fine for now. Please consider if you want to address that one new comment in this PR or the future one. I will leave it up to you.

Library/test-helpers/lib.sh
kkaarreell: I will deal with that in a separate commit later.

PR #366, opened 2023-04-14
Update exit codes according to keylime pr1337

kkaarreell:


Koncpa (approved):
LGTM!

kkaarreell:
I have removed the redirect. Tests will fail because the change is not yet in keylime upstream.

PR #371, opened 2023-04-19
Define SERVERS and CLIENTS based on TMT_ROLE_ variables

kkaarreell:
Define legacy (Beaker-like) environment variables `SERVERS` and `CLIENTS` if the respective role-based variables are defined by `tmt`. This change relates to the currently WIP multi-host test support for `tmt` developed in https://github.com/teemtee/tmt/pull/1790.


kkaarreell:
Should be backported to `rhel-` branches.

PR #374, opened 2023-04-20
Sync container support with the main branch

kkaarreell:


Library/test-helpers/lib.sh
```diff
@@ -1880,21 +1880,87 @@ limeconDeleteNetwork() {
     podman network rm -f $NAME
 }
 
+
```
Koncpa, line 1883: unnecessary new line?

Koncpa (changes requested):
Otherwise LGTM!

Library/test-helpers/lib.sh
kkaarreell: Yep.

Koncpa (approved):

PR #375, opened 2023-04-24
Enable Packit CI on all Fedoras

kkaarreell:
Fedora bug 2171376 should be fixed by now.


Koncpa:
The newest version of TPM tools still cannot be found and machine install only `tpm2-tools-5.5-1.fc38.x86_64`.

Koncpa (approved):

PR #376, opened 2023-04-24
When installing from Copr, set the measured boot log location

ansasaki:
If the TPM_BINARY_MEASUREMENTS environment variable is set, set the
measured boot log location when installing the agent from Copr.


kkaarreell:
Very good point, I was just wondering what's behind those failures in my other PR and it took me some time to realize that environment variables from `tmt` won't impact services started by systemd.

kkaarreell (approved):

kkaarreell:
The failure Is unrelated, we should extend maximum test time fór the failing test.

PR #378, opened 2023-04-26
Add previously removed limeconRunSystemd func

Koncpa:
Function provide start of systemd container.

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -2161,6 +2161,53 @@ limeconRunRegistrar() {
 true <<'=cut'
 =pod
 
+=head2 limeconRunSystemd
+
+Container run via podman with specified parameters.
+
+    limeconRunSystemd NAME TAG IP NETWORK EXTRA_PODMAN_ARGS
+
+=item NAME
+
+Set name of container.
+
+=item TAG
+
+Name of image tag.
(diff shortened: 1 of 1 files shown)


Koncpa:
Need to be cherry-pick also to rhel-9-main.

Library/test-helpers/lib.sh
```diff
@@ -2161,6 +2161,58 @@ limeconRunRegistrar() {
...
+limeconRunSystemd() {
+
+    local NAME=$1
+    local TAG=$2
+    local IP=$3
+    local NETWORK=$4
+    local COMMAND=$5
+    local EXTRA_PODMAN_ARGS=$6
+
+    limeconRun $NAME $TAG $IP $NETWORK $COMMAND $EXTRA_PODMAN_ARGS
```
kkaarreell, line 2209: I would expect you would be running `/sbin/init` here, instead of passing the command. And that there won't be any `COMMAND` in arguments.

kkaarreell (approved):
Assuming you have functionally tested IT. 😉

Koncpa:
Good catch @kkaarreell, actually. I plan to test it after backport to rhel-9-main, but It was need to be done before, thanks to that I found expanding variable error. Now it's properly tested.

PR #380, opened 2023-05-04
Ipv6 multihost test

Koncpa:


container/functional/keylime_ipv6_multihost/test.sh
```diff
@@ -0,0 +1,134 @@
...
+    rlPhaseEnd
+
+    rlPhaseStartTest "Add keylime agent"
+        rlRun -s "keylime_tenant -v $IP_VERIFIER  -t $IP_AGENT -u $AGENT_ID --runtime-policy policy.json -f /etc/hosts -c add"
+        rlRun "limeWaitForAgentStatus $AGENT_ID 'Get Quote'"
+        rlRun -s "keylime_tenant -c cvlist"
+        rlAssertGrep "{'code': 200, 'status': 'Success', 'results': {'uuids':.*'$AGENT_ID'" $rlRun_LOG -E
+    rlPhaseEnd
+
+    rlPhaseStartTest "Running allowed scripts should not affect attestation"
```
kkaarreell, line 100: I am thinking whether we should remove this and the following test phase. It would make sense to keep agent failure in case we would be also testing/using revocation notifications.. which would be probably also worth doing.

Library/test-helpers/lib.sh
```diff
@@ -1889,10 +1889,14 @@ true <<'=cut'
 
 Create container network.
 
-    limeconCreateNetwork NAME SUBNET
+    limeconCreateNetwork --ipv6 NAME SUBNET
```
kkaarreell, line 1892: Just a nitpick, it is common to list optional options and arguments as `[--ipv6]`.

kkaarreell (approved):
Nice, thanks a lot!

PR #384, opened 2023-05-15
Add conditional to clear tpm2 cache only on machines with HW/QEMU TPM

Koncpa:
IMA setup task could be used also on scenarios where is not used TPM, so tpm2_clear could be useless and just cause FAIL of setup task, added new conditional to check if task run on HW/QEMU TPM machine. 

Example of usage without TPM: [fapolicyd PR](https://github.com/RedHat-SP-Security/tests/pull/125/files#diff-c4ab242efdb76106e77b0c20a45b90bd44f8bd60e33b6bf3d41298c9d7037f8bR16)

### DIFF
--- a/setup/configure_kernel_ima_module/test.sh
+++ b/setup/configure_kernel_ima_module/test.sh
@@ -31,7 +31,7 @@ rlJournalStart
         # install IMA policy
         rlRun "limeInstallIMAConfig ${IMA_POLICY_FILE}"
         # clear TPM
-        if ! limeTPMEmulated; then
+        if ! limeTPMEmulated && [ -c /dev/tpmrm0 ]; then
             rlRun "tpm2_clear"
         fi
         # FIXME: workaround for issue https://github.com/keylime/keylime/issues/1025


Koncpa:
Manually tested on QEMU TPM machine and on machine without TPM.

setup/configure_kernel_ima_module/test.sh
```diff
@@ -31,7 +31,7 @@ rlJournalStart
         # install IMA policy
         rlRun "limeInstallIMAConfig ${IMA_POLICY_FILE}"
         # clear TPM
-        if ! limeTPMEmulated; then
+        if ! limeTPMEmulated && [ -c /dev/tpmrm0 ] || [ -c /dev/tpm0 ]; then
```
kkaarreell, line 34: I think just a check for one of them would be enough. In tests we rely on `/dev/tpmrm0` anyway.

Koncpa: Hmm that's true @kkaarreell , but do you think that `/dev/tpm0` is not used anymore? I read somewhere that it's used for older version of  TPM (1.2 or 2.0). I just think that could be useful also upgrade our tests to conditionally handle `/dev/tpm0` or `/dev/tpmrm0`.

kkaarreell: What I mean is that although both files are present, `/dev/tmp0' is for the direct device access while `/dev/tmprm0` it the access to TPM through a kernel resource manager and that is what we are using. In recent kernels there should always be also `/dev/tpmrm0` when `/dev/tmp0` exists. Btw, keylime won't work with TPM 1.2.

Koncpa: Ok, so I'll keep just `/dev/tpmrm0`.

PR #391, opened 2023-05-31
Export SERVERS and CLIENTS following tmt topology

kkaarreell:
`tmt` is changing the way how topology data are exported. This PR updates `sync` library so it works following `tmt` PR https://github.com/teemtee/tmt/pull/2072.

### DIFF
--- a/Library/sync/lib.sh
+++ b/Library/sync/lib.sh
@@ -52,9 +52,28 @@ the library load.
 
 # we are using hardcoded paths so they are preserved due to reboots
 export __INTERNAL_syncStatusFile=/var/tmp/sync-status
+
+# for backvards compatibility define SERVERS and CLIENTS variables using tmt topology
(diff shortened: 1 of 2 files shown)


kkaarreell:
@Koncpa FYI, I have just confirmed through manual testing that these updates are correct and work with the `tmt` version introducing this change.

kkaarreell:
We should backport this change to rhel-9-main branch at least, hopefully it would be possible.

Koncpa (approved):

kkaarreell:
> We should backport this change to rhel-9-main branch at least, hopefully it would be possible.

This is done in https://github.com/RedHat-SP-Security/keylime-tests/pull/393

kkaarreell:
F37 failure is unrelated. I will keep the failure here in order to enable investigation by Testing Farm folks.

PR #392, opened 2023-06-01
Unify .packit.yaml in the main branch

kkaarreell:
Since Packit RFE https://github.com/packit/packit-service/issues/2051 implemented we can unify .packit.yaml in the main branch so we won't have to customize it when rebasing tests to rhel-9-main or fedora-rawhide branch.


Koncpa:
Yeah, it seems like that.

PR #394, opened 2023-06-02
Retry when uploading files

kkaarreell:
It seems that previously added oshi.at is not that reliable, we are observing errors like
```
curl DBI connect failed: Too many connections at /usr/share/perl5/DBIx/Connector.pm
```
Therefore adding one additional file hosting webpage and adding support for verbose mode to simplify
future troubleshooting.


kkaarreell:
Hi @Koncpa , I am sorry but I should merge this one to fix coverage measurement for upstream.

PR #401, opened 2023-06-20
Remove workaround for multiple CA certificates

ansasaki:
Previously, the rust keylime agent supported only a single file in the `trusted_client_ca` option.  The support for a list in `trusted_client_ca` was added in https://github.com/keylime/rust-keylime/pull/604.


kkaarreell:
I see, this is still failing due to the problem with compiling the rust agent. I think it would be safe to merge this PR as the modified test is passing, still I would rather see all tests passing.

kkaarreell (approved):

PR #405, opened 2023-07-03
New scenario with keylime tenant in container

Koncpa:
New scenario with keylime tenant in container.
Add new limecon function which prepare, build
and manage tenant container.


Library/test-helpers/lib.sh
```diff
@@ -1166,32 +1178,44 @@ limeWaitForAgentStatus() {
...
 
     for I in `seq $TIMEOUT`; do
-        keylime_tenant -c status -u $UUID &> $OUTPUT
-	if grep -E -q "\"operational_state\": \"$STATUS\"" $OUTPUT; then
-            cat $OUTPUT
-	    rm $OUTPUT
-	    return 0
-	fi
-        sleep 1
+	    if [ -n "$NAME" ]; then
```
kkaarreell, line 1181: too big text indentation?

Library/test-helpers/lib.sh
```diff
@@ -1134,10 +1134,18 @@ true <<'=cut'
 Run 'keylime_tenant -c status' wrapper repeatedly up to TIMEOUT seconds
 until the expected agent status is returned.
 
-    limeWaitForAgentStatus UUID STATUS [TIMEOUT]
+    limeWaitForAgentStatus [--container] [NAME] UUID STATUS [TIMEOUT]
```
kkaarreell, line 1137: What I do not like about the `--container` option is that it could be misleading. One could also read it that the agent is running in a container. I would propose to use something like `--tenant-container`, even though it is longer. We could eventually support `-t` or `--tc` as a shortcut. Also, instead of variable `NAME` in the code I would also prefer something like `TENANT_CONTAINER` or `TENANT_CONT_NAME`. Just something to consider.

Library/test-helpers/lib.sh
kkaarreell, line 1181: Since you are not declaring `NAME` as `local` always, it could be non-empty if such a variable is defined by a user. Better always declare `local` variable (preferably with different name, more on that above).

Library/test-helpers/Dockerfile.tenant
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/centos/centos:stream9
+COPY id_*.pub /root/
+COPY lime_con_start.sh /usr/local/bin/lime_con_start
+RUN dnf install -y keylime-agent-rust util-linux-core openssh openssh-server which && \
+  dnf clean all && \
+  ssh-keygen -A && mkdir -p /root/.ssh && \
```
kkaarreell, line 6: My idea was that for tenant we would each time start a new container (i.e. the container won't be running as a server). From this perspective running ssh server there is probably not necessary.

kkaarreell: That also applies to lime_con_start script usage.

Library/test-helpers/lib.sh
```diff
@@ -1208,19 +1232,27 @@ Returns 0 when the start was successful, 1 otherwise.
...
     local UUID="$1"
     local OUTPUT=`mktemp`
     [ -z "$1" ] && return 3
     [ -n "$2" ] && TIMEOUT=$2
 
     for I in `seq $TIMEOUT`; do
-        keylime_tenant -c regstatus -u $UUID &> $OUTPUT
-	if grep -q "Agent $UUID exists on registrar" $OUTPUT; then
+        if [ -n "$NAME" ]; then
+            podman exec -i $NAME keylime_tenant -c regstatus -u $UUID &> $OUTPUT
```
kkaarreell, line 1246: I guess there is one thing we understand differently. In your case tenant container is a "service", a long-running process that is being used to query verifier/registrar eventually. 
In my understanding this is a one-shot container, that is started whenever we need to call keylime_tenant and ends rigth after it. IMHO there is no need to keep it running for a long time. 
This misunderstanding projects into the whole code. Probably we should discuss it in person.

Library/test-helpers/lib.sh
```diff
@@ -2217,6 +2249,50 @@ limeconRunSystemd() {
...
+=cut
+
+limeconRunTenant() {
+
+    local NAME=$1
+    local TAG=$2
+    local IP=$3
+    local NETWORK=$4
+
+    limeconRun $NAME $TAG $IP $NETWORK "/sbin/init" "--volume=/etc/keylime/:/etc/keylime/"
```
kkaarreell, line 2287: In my opinion this function should simply start the container, running the keylime_tenant command and passing the required arguments.. and finish after that. I.e. there won't be no cert copying, those would be mounted to the container when starting, and there won't be `podman exec`.

kkaarreell (changes requested):

Library/test-helpers/Dockerfile.tenant
```diff
@@ -0,0 +1,9 @@
+FROM quay.io/centos/centos:stream9
+COPY id_*.pub /root/
+COPY lime_con_start.sh /usr/local/bin/lime_con_start
+RUN dnf install -y keylime-agent-rust util-linux-core openssh openssh-server which && \
```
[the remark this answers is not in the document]
Koncpa: Probably I use agent docker file as model and I forgot to change it. Thank you for catch.

Library/test-helpers/lib.sh
Koncpa: Should be done for now.

Library/test-helpers/lib.sh
Koncpa: DONE.

Library/test-helpers/Dockerfile.tenant
Koncpa: DONE.

Library/test-helpers/lib.sh
```diff
@@ -1166,20 +1166,23 @@ limeWaitForAgentStatus() {
...
 
     for I in `seq $TIMEOUT`; do
-        keylime_tenant -c status -u $UUID &> $OUTPUT
-	if grep -E -q "\"operational_state\": \"$STATUS\"" $OUTPUT; then
+        if [ -n "$limeconKeylimeTenantCmd" ]; then
+            $limeconKeylimeTenantCmd "-c status -u $UUID" &> $OUTPUT
+        else
+            keylime_tenant -c status -u $UUID &> $OUTPUT
+        fi
+        if grep -E -q "\"operational_state\": \"$STATUS\"" $OUTPUT; then
```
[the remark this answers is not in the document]
Koncpa: Thanks for noticing.

Library/test-helpers/lib.sh
```diff
@@ -612,6 +612,38 @@ limeStopAgent() {
...
+
+=back
+
+Returns 0 when the stop was successful, non-zero otherwise.
+
+=cut
+
+limeKeylimeTenant() {
+
+    local TENANT_CMD=$1
```
kkaarreell, line 635: I think we could avoid using extra `"` around the cmd argument while calling `limeKeylimeTenant` by using `$@`, i.e. either pass `"$@" ` to the containerized tenant or `$@` to keylime_tenant.

Library/test-helpers/lib.sh
```diff
@@ -2217,6 +2248,63 @@ limeconRunSystemd() {
...
+    local TAG=$2
+    local IP=$3
+    local NETWORK=$4
+    local TENANT_CMD=$5
+    local MOUNT_TENANT="--volume=/etc/keylime/:/etc/keylime/"
+
+    if [ -d cv_ca ]; then
+        MOUNT_TENANT="--volume $PWD/cv_ca:/var/lib/keylime/cv_ca/:z $MOUNT_TENANT"
+    fi
+    if [ -f "policy.json" ]; then
```
kkaarreell, line 2295: I think we should be rather mounting/passing a specific directory that would be referenced from the inside of the container. This is not just about policy.json, we could also pass while directory with a payload.

Library/test-helpers/lib.sh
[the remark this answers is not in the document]
kkaarreell: Just merged it, please rebase And resolve the conflict.

Library/test-helpers/lib.sh
```diff
@@ -2219,6 +2250,50 @@ limeconRunSystemd() {
...
+    local PODMAN_CMD=$1
+    local TENANT_CMD=$2
+    local MOUNT_DIR=$3
+    local MOUNT_TENANT="--volume=/etc/keylime/:/etc/keylime/"
+
+    if [ -d cv_ca ]; then
+        MOUNT_TENANT="--volume $PWD/cv_ca:/var/lib/keylime/cv_ca/:z $MOUNT_TENANT"
+    fi
+    
+    rlRun "podman run $MOUNT_DIR $MOUNT_TENANT $PODMAN_CMD keylime_tenant $TENANT_CMD"
```
kkaarreell, line 2288: There should not be rlRun

container/functional/keylime_all_in_container-basic-attestation/test.sh
```diff
@@ -0,0 +1,159 @@
...
+        rlRun "podman cp $CONT_VERIFIER:/var/lib/keylime/cv_ca/ ."
+
+        #run registrar container
+        CONT_REGISTRAR="registrar_container"
+        rlRun "limeconRunRegistrar $CONT_REGISTRAR $TAG_REGISTRAR $IP_REGISTRAR $CONT_NETWORK_NAME keylime_registrar /etc/keylime $(realpath ./cv_ca)"
+        rlRun "limeWaitForRegistrar 8891 $IP_REGISTRAR"
+
+        CONT_TENANT="tenant_container"
+        # define limeconKeylimeTenantCmd so that the keylime container can be used by limeWaitForAgentStatus etc.
+        limeconKeylimeTenantCmd="--name $CONT_TENANT --net $CONT_NETWORK_NAME --ip $IP_TENANT localhost/$TAG_TENANT"
```
kkaarreell, line 96: Any specific reason why ` limeconRunTenant` is not part of this command already? It seems confusing to call it command but store only options. I would suggest to include ` limeconRunTenant` here as well.

Koncpa: Done.

PR #420, opened 2023-08-02
Switch faked_measured_boot_log context values to yes/no

kkaarreell:


kkaarreell:
@Koncpa FYI, I have also added a commit updating .packit.yaml. This is something we would need to do with every Fedora release.

Koncpa (approved):

PR #426, opened 2023-08-04
Use development C9S repos in .packit.yaml

kkaarreell:
Commonly used CentOS Stream 9 mirrors are one week behind the main repository. Therefore let's use the main repo instead of mirrors.


Koncpa (approved):

kkaarreell:
`/functional/basic-attestation-on-localhost` passed which means that the change works. The rest would be failing on container tests which is addressed in another PR.

PR #429, opened 2023-08-04
Test upsteam container images

ansasaki:
- Add a helper function to pull images from a registry.
- Add a test plan to test upstream container images
- Use `--entrypoint` to override the container entrypoint and allow passing arguments to the commands inside the container in `limeconRun`
- Allow passing the agent working directory in `limeconRunAgent`
- Add a Dockerfile for the revocation webhook in `keylime_ipv6_multihost` test
- Avoid the need for python inside the container in `keylime_ipv6_multihost` by replacing python payload script with shell script


plans/upstream-keylime-containers.fmf
```diff
@@ -0,0 +1,33 @@
...
+
+execute:
+    how: tmt
+
+adjust+:
+  - when: target_PR_branch is defined and target_PR_branch != main
+    enabled: false
+    because: we want to run this plan only for PRs targeting the main branch
+
+  - when: distro != fedora-37
```
kkaarreell, line 32: FYI, it is not possible to configure swtpm device on C9S. I am just updating the other plan in
https://github.com/RedHat-SP-Security/keylime-tests/pull/436

plans/upstream-keylime-containers.fmf
```diff
@@ -0,0 +1,33 @@
+summary:
+  Tests used by Packit/TFT CI on Github to test upstream keylime containers
+
+environment+:
+  TPM_BINARY_MEASUREMENTS: /var/tmp/binary_bios_measurements
+  KEYLIME_RUST_CODE_COVERAGE: 1
```
kkaarreell, line 6: For containers we are not really measuring code coverage so we probably don't have to set this one.

plans/upstream-keylime-containers.fmf
```diff
@@ -0,0 +1,33 @@
...
+  AGENT_DOCKERFILE: Dockerfile.upstream.c9s
```
kkaarreell, line 7: Why do we define dockerfile for agent and not for other services? Do we need dockerfile definitions at all, given we define images?

kkaarreell:
I am wondering what's behind those test failure, seems like there was an issue with agent start. Unfortunately, logging for containers is not currently ideal.

plans/upstream-keylime-containers.fmf
ansasaki: Thanks, I'll fix.

plans/upstream-keylime-containers.fmf
ansasaki: Thank you, I'll fix this

ansasaki:
> I am wondering what's behind those test failure, seems like there was an issue with agent start. Unfortunately, logging for containers is not currently ideal.
> 
> EDIT: seems there is in fact something in logs
> 
> ```
> error: unexpected argument 'keylime_agent' found
> 
> Usage: sudo RUST_LOG=keylime_agent=trace ./target/debug/keylime_agent
> 
> For more information, try '--help'.
> ```

I already fixed this locally. For images that already have an entrypoint set, you need to override it, otherwise the trailing commands are provided as arguments.

But I hit another issue: the upstream agent container image does not create an internal `keylime` user, making it impossible to drop privileges. Either we run as root or we add a `keylime` user there.

I think it is better to add the `keylime` user there for now. I'll create a PR and ask for review.

ansasaki:
Added the `keylime` user via https://github.com/keylime/rust-keylime/pull/635

kkaarreell:
I am trying to address C9S failure in https://github.com/RedHat-SP-Security/keylime-tests/pull/450

container/functional/keylime_agent_container-basic-attestation/test.sh
```diff
@@ -8,6 +8,8 @@
 
 [ -n "$AGENT_DOCKERFILE" ] || AGENT_DOCKERFILE=Dockerfile.upstream.c9s
 
+[ -n "$REGISTRY" ] || REGISTRY=quay.io
```
kkaarreell, line 11: Should we make images configurable through envvars as well?

container/functional/keylime_ipv6_multihost/payload-script/local_action_modify_payload.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+TYPE=$(grep -o "\"type\":[ ]*\"\([^\"]*\)\"" "$1" | sed -e 's/\"type\":[ ]*\"\([^\"]*\)\"/\1/g')
+
+if [ "$TYPE" != "revocation" ]; then
+    echo "Input JSON type is not revocation"
+    exit 1
+fi
+
+EVENT_UUID=$(grep -o "\"agent_id\":[ ]*\"\([^\"]*\)\"" "$1" | sed -e 's/\"agent_id\":[ ]*\"\([^\"]*\)\"/\1/g')
```
kkaarreell, line 17: Just to confirm, you are not using `jq` because it may not be available in the keylime image, am I right?

container/functional/keylime_agent_container-basic-attestation/test.sh
kkaarreell: Now I can see we do configure it through envvars, it's just not clearly visible in the test. What about adding a comment to the top?

kkaarreell (approved):
Thank you for the update. Looks good, I would just consider adding a comment to a test mentioning which images can be defined using envvars from the outside.

kkaarreell:
@Koncpa please take a look at it as it introduces some new functions and concepts.

Koncpa:
> @Koncpa please take a look at it as it introduces some new functions and concepts.

Sure, I'll take a look.

Library/test-helpers/lib.sh
```diff
@@ -2041,11 +2053,69 @@ limeconPrepareImage() {
...
+
+Pull the requested image from the repository and tag with the provided tag.
+
+    limeconPullImage REGISTRY IMAGE [LOCAL_NAME_TAG]
+
+=over
+
+=item REGISTRY
+
+The registry from where the image will be pulled
```
Koncpa, line 2066: Missing dot.

container/functional/keylime_ipv6_multihost/payload-script/local_action_modify_payload.sh
ansasaki: Right, the goal was to require very minimum dependencies to run.

Koncpa:
Just one nitpick, but otherwise LGTM! Thank you.

Library/test-helpers/lib.sh
ansasaki: Thanks!

PR #430, opened 2023-08-04
Added forgotten mktemp

kkaarreell:


kkaarreell:
Verified the fix with local testing.

PR #435, opened 2023-08-05
Backport fixes from rhel-9-main

kkaarreell:
Backport `.packit.yaml` update and distribution testplan updates from `rhel-9-main` so that they won't be lost with a future rewrite.


kkaarreell:
Test failure on F37 is unrelated to these changes but it is worth investigating.

PR #436, opened 2023-08-10
Enable upstream-keylime-swtpm-dev for C9S

kkaarreell:
vtpm_tpm_proxy module is finally available in C9S and we can start using it.

### DIFF
--- a/plans/upstream-keylime-swtpm-dev.fmf
+++ b/plans/upstream-keylime-swtpm-dev.fmf
@@ -28,5 +28,5 @@ adjust+:
     enabled: false
     because: we want to run this plan only for PRs targeting the main branch
 
-  - when: distro != fedora-37
+  - when: distro != fedora-37 and distro != centos-stream-9
     enabled: false


Koncpa (approved):

Library/test-helpers/lib.sh
```diff
@@ -49,6 +48,7 @@ The library provides shell function to ease keylime test implementation.
 #   Variables
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+
```
Koncpa, line 51: leftover?

Koncpa (approved):

Library/test-helpers/lib.sh
kkaarreell: fixed

ansasaki (approved):
LGTM!

PR #437, opened 2023-08-10
Set keylime_port_t label for random ports use in tests

Koncpa:


kkaarreell (approved):

kkaarreell:
@Koncpa How to approach this for the main branch? We do not want to loose it but I assume it won't work on various Fedoras, right? Maybe we could do the change conditionally inside the test. WDYT?

Koncpa:
Sure, it is possible do it conditionally. It will be work just for RHEL-9.3.0, where are keylime ports labeled.

Koncpa:
But do we want to add `if its RHEL` conditions into test?

PR #443, opened 2023-08-15
Fix check for TMT_TOPOLOGY_BASH

kkaarreell:
backporting the fix that is already used in main and rhel-9-main


kkaarreell:
Packit has an outage. Since this change is small and already used in tho other branches I will take the liberty and merge it.

PR #447, opened 2023-08-15
Add allowlist support to multihost test

kkaarreell:
This is necessary for interoperability testing with older keylime


kkaarreell:
Tested locally.

PR #457, opened 2023-08-25
compatibility: Add test for agent API bump

ansasaki:
When the agent updates its API version while being monitored by a verifier, the verifier will automatically update the used API version.

See: https://github.com/keylime/keylime/issues/1457


compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
```diff
@@ -0,0 +1,112 @@
...
+            rlRun "limeStartIMAEmulator"
+        fi
+        sleep 5
+        # start keylime_verifier
+        rlRun "limeStartVerifier"
+        rlRun "limeWaitForVerifier"
+        rlRun "limeStartRegistrar"
+        rlRun "limeWaitForRegistrar"
+        # create allowlist and excludelist
+        rlRun "limeCreateTestPolicy ${TESTDIR}/*"
```
kkaarreell, line 44: remove the `/*` part to fix the failure.

kkaarreell: In fact it seems that `TESTDIR` is not define and therefore you can completely remove this parameter.

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
```diff
@@ -0,0 +1,112 @@
...
+        rlRun "limeStopVerifier"
+        if limeTPMEmulated; then
+            rlRun "limeStopIMAEmulator"
+            rlRun "limeStopTPMEmulator"
+            rlRun "limeCondStopAbrmd"
+        fi
+        limeSubmitCommonLogs
+        limeClearData
+        limeRestoreConfig
+        limeExtendNextExcludelist "$TESTDIR"
```
kkaarreell, line 108: remove, not defined anyway.

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
```diff
@@ -0,0 +1,112 @@
...
+
+        WORKDIR=$( mktemp -d -p "/var/tmp" )
+    rlPhaseEnd
+
+    rlPhaseStartTest "Compile old and new keylime agent"
+        rlRun "git clone ${RUST_KEYLIME_UPSTREAM_URL} ${WORKDIR}/rust-keylime"
+        rlRun "pushd ${WORKDIR}/rust-keylime"
+        rlRun "git checkout v0.2.1"
+        rlRun "cargo build && cp ./target/debug/keylime_agent ${WORKDIR}/keylime_agent-2.1"
+        rlRun "git checkout ${RUST_KEYLIME_UPSTREAM_BRANCH}"
```
kkaarreell, line 54: Is it necessary to build the "current" version again? Agent built from PR sources should be already built by the former `/setup/` task so we should simply use it, or not?

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
```diff
@@ -0,0 +1,112 @@
...
+_EOF"
+        rlRun "expect script.expect"
+        rlRun "limeWaitForAgentStatus $AGENT_ID 'Get Quote'"
+        rlRun -s "keylime_tenant -c cvlist"
+        rlAssertGrep "{'code': 200, 'status': 'Success', 'results': {'uuids':.*'$AGENT_ID'" $rlRun_LOG -E
+    rlPhaseEnd
+
+    rlPhaseStartTest "Verify that API version is automatically bumped"
+        rlRun "kill $AGENT_PID"
+        rlRun "RUST_LOG=keylime_agent=debug ${WORKDIR}/keylime_agent >> $(limeAgentLogfile) 2>&1 &"
```
kkaarreell, line 77: Can't we simply start the service?

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
ansasaki: OK!

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
ansasaki: I wasn't sure if the current version would be always available. It would be great if we can avoid recompiling it again.

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
ansasaki: If we don't recompile again and use the installed keylime, then I think we can simply start the service

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
kkaarreell: There will be one installed, otherwise all tests would fail.

ansasaki:
@Koncpa @kkaarreell Tests passed in https://github.com/keylime/keylime/pull/1459 using this branch

kkaarreell (approved):

PR #468, opened 2023-08-30
Use latest c9s repos also for Dockerfiles

kkaarreell:


Koncpa (approved):

kkaarreell:
@ansasaki Hi, just FYI, I had to do small changes in the container related code because the rewrite broke tests that copied ssh keys to a container.

PR #472, opened 2023-08-31
Set default ref for fedora-39 for dynamic ref

Koncpa:


.tmt/dynamic_ref.fmf
```diff
@@ -20,6 +20,9 @@ adjust:
  - when: distro = fedora-38
    ref: fedora-38
    continue: false
+ - when: distro = fedora-39
+   ref: fedora-39
+   continue: false
  - when: distro = fedora
    ref: fedora-main
```
kkaarreell, line 27: change this to `fedora-rawhide`
And before that also add a rule for F38
```
when: distro = fedora-38
ref: fedora-main
continue: false
```

Koncpa: I add rule for f37 and specify, that fedora-main will be applied for fedora-rawhide

kkaarreell (approved):

PR #490, opened 2023-09-25
Set label for ports only if are not yet set

Koncpa:
Setting non default ports for keylime requires
additional SELinux labeling for used ports.
When test is rerun in same enviroment ports are
not needed set again and setup could cause failures in test.

### DIFF
--- a/functional/keylime-non-default-ports/test.sh
+++ b/functional/keylime-non-default-ports/test.sh
@@ -110,6 +110,14 @@ _EOF"
         rlRun "limeStopAgent"
         rlRun "limeStopRegistrar"
         rlRun "limeStopVerifier"
+        #remove keylime_port_t label from non default ports
+        if rlIsRHEL '>=9.3' || rlIsFedora '>=38' || rlIsCentOS '>=9';then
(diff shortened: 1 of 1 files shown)


kkaarreell (approved):

functional/keylime-non-default-ports/test.sh
```diff
@@ -110,6 +110,14 @@ _EOF"
         rlRun "limeStopAgent"
         rlRun "limeStopRegistrar"
         rlRun "limeStopVerifier"
+        #remove keylime_port_t label from non default ports
+        if rlIsRHEL '>=9.3' || rlIsFedora '>=38' || rlIsCentOS '>=9';then
+            rlRun "semanage port -a -t keylime_port_t -p tcp 19002"
```
kkaarreell, line 115: Shouldn't be there `--remove` instead of `--add`?

kkaarreell (changes requested):
I believe there should be `semanage port --delete ...`

kkaarreell (approved):

kkaarreell:
@Koncpa those test failures are due to the recently merged https://github.com/keylime/keylime/pull/1480
We need to update our test in `main` accordingly.

PR #491, opened 2023-09-25
Set label for ports only if are not yet set

Koncpa:
Setting non default ports for keylime requires
additional SELinux labeling for used ports.
When test is rerun in same enviroment ports are
not needed set again and setup could cause failures in test.

### DIFF
--- a/functional/keylime-non-default-ports/test.sh
+++ b/functional/keylime-non-default-ports/test.sh
@@ -110,6 +110,14 @@ _EOF"
         rlRun "limeStopAgent"
         rlRun "limeStopRegistrar"
         rlRun "limeStopVerifier"
+        #remove keylime_port_t label from non default ports
+        if rlIsRHEL '>=9.3' || rlIsFedora '>=38' || rlIsCentOS '>=9';then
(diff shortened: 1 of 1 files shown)


functional/keylime-non-default-ports/test.sh
```diff
@@ -110,6 +110,8 @@ _EOF"
         rlRun "limeStopAgent"
         rlRun "limeStopRegistrar"
         rlRun "limeStopVerifier"
+        #remove keylime_port_t label from non default ports
+        rlRun "semanage port -D"
```
kkaarreell, line 114: Can we remove just labels from ports we have configured? I am afraid that doing this could possible collide with any a different system setup we might do.

Koncpa: Sure, I changed that..

kkaarreell (approved):

kkaarreell (changes requested):
I believe the semanage command uses incorrect arguments.

kkaarreell (approved):
Sorry, that was the other PR. This one is OK.

PR #499, opened 2023-10-10
install_upstream_rust_keylime: allow setting measurements location

ansasaki:
Use the `testing` feature and set the TPM_BINARY_MEASUREMENTS
environment variable for the agent service when the
TPM_BINARY_MEASUREMENTS environment variable is set.

This avoids the need for modifying the code to change the location of the binary measurements log.


galmasi (approved):
LGTM. I hope this will straighten out https://github.com/keylime/rust-keylime/pull/665.

kkaarreell (approved):
Thank you very much.

PR #501, opened 2023-10-17
Adjust test to the keylime_create_policy upstream

kkaarreell:
Cherry-pick from the main branch.


kkaarreell:
The modified test is expected to fail as the fix is not yet released.

PR #505, opened 2023-10-31
Add cve test to test suite

Koncpa:


Library/test-helpers/lib.sh
```diff
@@ -2356,7 +2356,9 @@ limeconRunAgent() {
         # On rootless container, this could be done with 'podman unshare'
         podman run --rm --attach stdout $EXTRA_ARGS --entrypoint chown localhost/agent_image -R keylime:keylime /var/lib/keylime
     else
-        EXTRA_ARGS="--tmpfs /var/lib/keylime/ $EXTRA_ARGS"
+        # Find out better way to handle this: upstream containers have /var/lib/keylime owned by root, also --tmpfs mount changes
+        # directory ownership to root. So we need to set mode=777 to let agent to write there.
+        EXTRA_ARGS="--mount=type=tmpfs,dst=/var/lib/keylime/,tmpfs-mode=777 $EXTRA_ARGS"
```
ansasaki, line 2361: I think this is OK for now, but we should remove this once the upstream container is fixed to have the correct ownership.

setup/configure_tpm_emulator/swtpm_permissive.te
```diff
@@ -13,6 +13,7 @@ permissive swtpm_t;
 init_daemon_domain(swtpm_t, swtpm_exec_t);
 
 dontaudit swtpm_t self:tcp_socket { accept listen };
+dontaudit swtpm_t swtpm_t:capability { sys_admin };
```
ansasaki, line 16: I'm not sure about the impact of this change. @Koncpa Could you please check this?

kkaarreell: For swtpm we have a custom SELinux module that runs swtpm in a permissive domain and we don't audit all AVCs. The reason for that is basically that our swtpm usage is slightly different from the general perception. While starting TPM on a running system is possible, the major use case for swtpm is rather emulating TPM in virtual systems and the existing SELinux policy follows that approach.

kkaarreell:
Results looks good, merging.

PR #511, opened 2023-11-03
Enable regression test in the Fedora distribution test plan

kkaarreell:


kkaarreell:
This test failure is expected, a new build with the fix is in progress.

PR #515, opened 2023-11-15
Use --noautoremove when removing installed keylime

kkaarreell:
While removing keylime yum has decided to remove some packages that were required by tests.


Koncpa (approved):

kkaarreell:
A test failure on Rawhide is unrelated to this update. I suspect this is another measured_boot_log parsing issue.

PR #527, opened 2023-12-01
AVC's workaround

Koncpa:


kkaarreell:
@Koncpa I have noticed in the latest CI runs that AVCs are gone. Maybe it has been fixed in the meantime.

Koncpa:
@kkaarreell Yeah, it's seems like that is fixed. On recent [PR](https://github.com/RedHat-SP-Security/keylime-tests/pull/531) don't see any similar AVC's denials.

PR #531, opened 2023-12-04
Fix dynref syntax and add rhel-8-main rule

kkaarreell:


kkaarreell:
Test failure on Rawhide is expected and unrelated.

Koncpa (approved):

PR #536, opened 2024-01-09
Fix error in MY_IP assignment

kkaarreell:


kkaarreell:
The failure in ` /functional/iak-idevid-register-with-certificates ` is caused by wget2 package which has replaced wget in Rawhide.

kkaarreell:
The other test failure is caused by https://github.com/keylime/keylime/issues/1496

Koncpa (approved):

PR #544, opened 2024-01-25
Wait for system being online after reboot

kkaarreell:


kkaarreell:
As discussed on `#tmt` Slack channel this change should not make a difference. But maybe we can merge it anyway to make the check explicit. WDYT?

Koncpa (approved):

PR #550, opened 2024-02-07
Workaround ahash compilation issue with nightly rustc

ansasaki:
On `compatibility/basic-attestation-on-localhost-api-version-bump`:

- Avoid compilation error when using nightly by updating ahash crate to version 0.8.7


kkaarreell (approved):
Thank you for the fix. Feel free to merge when you are comfortable.
Do we need to backport it to other test branches?

ansasaki:
> Thank you for the fix. Feel free to merge when you are comfortable. Do we need to backport it to other test branches?

Yes, probably we'll need backports. I will merge this to make the tests to pass on the open PRs

PR #557, opened 2024-03-15
Adjust XTRA due to changes in TF

kkaarreell:


kkaarreell:
I am merging this one since we need to test multihost tests and this change only impacts multihost tests.. Failures with agent building are unrelated.

PR #560, opened 2024-03-15
install_upstream_rust_keylime: Pin rust compiler version 1.74.1

ansasaki:
Replace the nightly version usage with a pinned version (1.74.1) that supports code instrumentation for test coverage measurement by `tarpaulin`


kkaarreell (approved):
Thank you. Test failures are unrelated, though we should also investigate it. Probably some change in upstream keylime that we didn't catch.

PR #580, opened 2024-05-08
change IDevID grep phrase check

Isaac-Matthews:
With the new [patch to enable handles](https://github.com/keylime/rust-keylime/pull/785) the phrasing here has changed to be less ambiguous.


kkaarreell:
Those failures are currently expected, unfortunately. I am discussing it with the Testing Farm team.

PR #582, opened 2024-05-14
Use version 0.2.5 of rust keylimer-agent

mpeters:


kkaarreell:
Hi Michael, the test has been fixed in https://github.com/RedHat-SP-Security/keylime-tests/pull/581 .

PR #583, opened 2024-05-16
Use the latest old supported version that is not the current version to test API bump

ansasaki:
This change is to cover a corner case when the python components API version is bumped before the agent API version is bumped.

In such a case, the latest old supported API version for the verifier would be the current agent API version. Then it is necessary to use the second to last to cause a difference in the API version.


kkaarreell (approved):

ansasaki:
The failures in Rawhide look unrelated, probably caused by some change in selinux (AVCs).

Maybe @Koncpa could take a look? In my experience this kind of issue in Rawhide has high chances of vanishing "automagically"

kkaarreell:
Rawhide AVC is unrelated, reported as https://bugzilla.redhat.com/show_bug.cgi?id=2280358

PR #589, opened 2024-06-12
Ignore if rpm-plugin-ima cannot be uninstalled

kkaarreell:


kkaarreell:
Rawhide AVCs are unrelated and already reported.

sergio-correia (approved):

kkaarreell:
AVCs on C10S are also unrelated and reported.

PR #592, opened 2024-06-21
Support named measured boot policies

niteeshkd:
Tests the support of the named measure boot policies as mentioned in Keylime [PR#1547](https://github.com/keylime/keylime/pull/1547) .


niteeshkd:
> /packit retest-failed

@ansasaki, is the failure because of the changes by this PR?

ansasaki:
> > /packit retest-failed
> 
> @ansasaki, is the failure because of the changes by this PR?

I don't think so, we are currently investigating it

ansasaki:
@kkaarreell Could you please review this? I don't think the errors on the tests are caused by the changes in this PR. Probably the failing tests can be waived and we can merge this.

I asked @Koncpa for help to identify the correct way to fix these AVCs on Fedora Rawhide and Centos Stream 10

kkaarreell:
Correct, those AVCs are not related to this change. I think it is safe to fix it. Thank you for improving the testsuite.

@Koncpa I am not sure for how long those AVCs tracked under https://issues.redhat.com/browse/RHEL-37631 will keep appearing, maybe we could temprorarily dontaudit them, WDYT?

kkaarreell (approved):
LGTM.

PR #595, opened 2024-07-16
Stop IDevID test being dependant on a previous test running

Isaac-Matthews:
At the moment, the IDevID test (```iak-idevid-register-with-certificates```) works because the test ```ek-cert-use-ek_handle-custom-ca_certs``` previously ran and created the tpm certs folder.

This change makes sure the folder has been created, allowing the test to be run independently of other tests running.


kkaarreell (approved):

kkaarreell:
Test failures are unrelated to this change, merging. Thank you for the fix.

PR #597, opened 2024-07-17
Move to c10s based dockerfiles

kkaarreell:
Since we are no longer able to install upstream keylime to c9s we should also base our container tests and dockerfiles on c10s.


kkaarreell:
It looks good. The Rawhide failure is a know rpm issue.

PR #601, opened 2024-07-17
Persisted and protected IDevID and IAK test

### LINKED ISSUE #587: Add a test to cover IDevID/IAK usage with passwords
The goal is to cover the changes introduced by https://github.com/keylime/rust-keylime/pull/785

@Isaac-Matthews It would be great if you could contribute a test case

Isaac-Matthews:
Fixes #587 
This is a test for IDevIDs and IAKs that are persisted in the TPM and are accessed via handle rather than regenerated. These keys may also be protected by password, this is also tested, as is the change to enable hex passwords for the TPM.

Relevant PRs:
[agent: enable persisted and protected IDevIDs](https://github.com/keylime/rust-keylime/pull/785), [agent: enable hex passwords for the TPM](https://github.com/keylime/rust-keylime/pull/769), [server: add config options](https://github.com/keylime/keylime/commit/7b3ea56aae40dc04ae4f28f0147486f90ff547fa)

### DIFF
--- a/functional/iak-idevid-persisted-and-protected/test.sh
+++ b/functional/iak-idevid-persisted-and-protected/test.sh
@@ -1,0 +1,169 @@
+#!/bin/bash
+# vim: dict+=/usr/share/beakerlib/dictionary.vim cpt=.,w,b,u,t,i,k
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+AGENT_ID="d432fbb3-d2f1-4a97-9ef7-75bd81c00000"
+TPM2_OPENSSL="https://github.com/tpm2-software/tpm2-openssl/releases/download/1.2.0/tpm2-openssl-1.2.0.tar.gz"
+CA_PWORD="keylimeca"
+CERT_DIR="/var/lib/keylime"
+TPM_CERTS="/var/lib/keylime/tpm_cert_store"
+
+rlJournalStart
+
+    rlPhaseStartSetup "Do the keylime setup"
+        rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
+        rlAssertRpm keylime
+        # update /etc/keylime.conf
+        limeBackupConfig
+        rlRun "TESTDIR=$(pwd)"
+        rlRun "TMPDIR=\$(mktemp -d)"
+        rlRun "pushd ${TMPDIR}"
+        # tenant, set to true to verify ek on TPM
+        rlRun "limeUpdateConf agent enable_iak_idevid true"
+        rlRun "limeUpdateConf registrar tpm_identity iak_idevid"
+        # if TPM emulator is present
+        if limeTPMEmulated; then
+            # start tpm emulator
+            rlRun "limeStartTPMEmulator"
+            rlRun "limeWaitForTPMEmulator"
+            rlRun "limeCondStartAbrmd"
+        fi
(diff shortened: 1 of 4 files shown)


kkaarreell:
Please rebase, that should help to get rid of most of test failures. Thank you.

functional/iak-idevid-persisted-and-protected/test.sh
```diff
@@ -0,0 +1,169 @@
...
+            rlRun "cd tpm2-openssl-1.2.0"
+            rlRun "./configure"
+            rlRun "make"
+            rlRun "make install"
+            #rlRun "make check"
+            rlRun "cd .."
+        rlPhaseEnd
+    fi
+
+    rlPhaseStartSetup "Create CA"
```
kkaarreell, line 52: Usually, we use [openssl-certgen beakerlib library](https://pagure.io/beakerlib-libraries/blob/master/f/openssl/certgen/lib.sh) to generate certificates, e.g. https://github.com/RedHat-SP-Security/keylime-tests/blob/7cad5f9dbb5bb1fb365b2a418ec75feba4f51d51/functional/basic-attestation-with-custom-certificates/test.sh (plus a require item in main.fmf)

Are there "ordinary" certificates or is there anything "special" about them? I am not strongly against doing it this way, even though I believe the other approach would be shorter..  But in cases these certs have any special property it would be good to mention it in a comment.

kkaarreell (approved):

functional/iak-idevid-persisted-and-protected/test.sh
kkaarreell: Hi @ansasaki , could you please review it as well?

ansasaki: Yes, I'll review

ansasaki: Sorry for taking long to get to review this.

I agree with what Karel wrote, I believe it would be quite simple to use the `certgen/certgen` beaker library to generate the certificates including the required SAN without having to carry the CA configuration files.

In my opinion it would clarify what is important to have in the certificate.

If you disagree or cannot make the changes, I think it is acceptable to keep the configuration files for now.

The only change I would like to request is to provide a better commit message.

Isaac-Matthews: No problem, thanks for the review. I will change the commit message.

For the certs - I don't disagree. However can I suggest we merge this first, and then when I have some time I will look into changing over to using certgen. This would also mean that the change can affect both this test and the other IDevID test as swapping the cert generation mechanism should probably be part of the same commit (the other test currently uses the same process that is used here).

kkaarreell:
Hi @Isaac-Matthews , it seems that all the related changes are already merged upstream. Can we merge this test PR as well?
FYI, in order to run this test in keylime CI we need to add it to the packit-ci.fmf plan in the upstream project.

Isaac-Matthews:
> Hi @Isaac-Matthews , it seems that all the related changes are already merged upstream. Can we merge this test PR as well? FYI, in order to run this test in keylime CI we need to add it to the packit-ci.fmf plan in the upstream project.

@kkaarreell Yes this is ready to be merged, I was just waiting for reviews. So please merge whenever.
For the plans I already have draft prs [1](https://github.com/keylime/rust-keylime/pull/814) [2](https://github.com/keylime/keylime/pull/1589) to enable the test. If any other changes are needed let me know.

PR #603, opened 2024-07-30
Add extra dependencies for the tests of enhancement-110 PR

sergio-correia:


kkaarreell (approved):

kkaarreell:
These are quite common deps, we could add them to main.fmf to `require` section.

sergio-correia:
> These are quite common deps, we could add them to main.fmf to `require` section.

Good point, and one of them was already present there. I updated it as per your suggestion.

kkaarreell (approved):

PR #604, opened 2024-08-02
Fix multi-host test for C10S

kkaarreell:


kkaarreell:
multihost test passing.
OK

PR #606, opened 2024-08-06
Fix typo in requires

kkaarreell:


kkaarreell:
The failure in /functional/keylime_tenant-ima-signature-sanity
is caused by regression in Rawhide python3.13
Filed as https://bugzilla.redhat.com/show_bug.cgi?id=2303107

Most likely the same applies to the upstream test failure.

kkaarreell:
I just realized we do not want this change.

PR #608, opened 2024-08-06
Make payload compatible with older keylime

kkaarreell:


kkaarreell:
rawhide test failures are known issues in other packages.

PR #634, opened 2024-08-20
Backport changes to use TLS on revocation notification webhook

ansasaki:
Backport the changes from `main` to use TLS on revocation notification webhook.

This is necessary to make tests on https://gitlab.com/redhat/centos-stream/rpms/keylime/-/merge_requests/37 to pass.


kkaarreell:
Passed against downstream

PR #635, opened 2024-08-26
Use separate CA for webhook server certificate

kkaarreell:


kkaarreell:
Not needed as we have https://github.com/RedHat-SP-Security/keylime-tests/pull/635

PR #637, opened 2024-08-27
restart journald

kkaarreell:
This is a "workaround" for one weird journald issue we are observing, causing frequent failures of this test.


kkaarreell:
Although we are not getting "all green",  I am afraid we need to merge this since the situation without these patches would be even worse.

PR #639, opened 2024-09-05
Modify keylime-tests to support SecureBoot on RHEL-10

kkaarreell:


kkaarreell:
I have filed
https://issues.redhat.com/browse/RHEL-57720
regarding the container test failure. Will switch to c10s image to bypass that.

PR #652, opened 2024-09-16
keylime_policy-commands: Do not expect algorithm in policy

ansasaki:
Previously, the keylime_policy command would set the IMA template hash algorithm to follow the algorithm set through --algo option.

This was an incorrect behavior and was recently fixed. Now the IMA template hash algorithm is hardcoded as SHA-1, following kernel implementation.


kkaarreell (approved):
Thank you for the update.
Multi-host test failures are unrelated, this is something I need to fix.

ansasaki:
I just merged the changes needed in `keylime/keylime`; I'll drop the temporary commit.

PR #657, opened 2024-09-24
Adjust tests to conform with upstream PR#1648

sergio-correia:
This PR adjusts the tests to conform with the changes in https://github.com/keylime/keylime/pull/1648:
1. the tool supports `--allowlist`/`--excludelist` instead of `--allowlist`/`--exclude-list`
2. the tool drops some confusing "short" options, like `-i` and `-A`, and keep only long versions instead, for those
3. the tool drops the superfluous option `--use-ima-measurement-list`, as we already have `--ima-measurement-list`; the behavior is the following: if `--ima-measurement-list` is used without any argument, it will use the default IMA measurement list file, `/sys/kernel/security/ima/ascii_runtime_measurements`; if a parameter is given as argument, that parameter indicates the file to read as a source for the IMA measurement list
4. finally, the tool itself is renamed to `keylime-policy` instead of `keylime_policy`


ansasaki (approved):
Just a reminder that this PR modified the name of a test case from `keylime_policy-commands` to `keylime-policy-commands` which will affect https://github.com/keylime/keylime/pull/1650

kkaarreell:
Hi @sergio-correia , could you please point keylime sources to your PR repo&branch so we can see here how it goes?
https://github.com/RedHat-SP-Security/keylime-tests/blob/main/plans/main.fmf#L8

kkaarreell (approved):

sergio-correia:
> Hi @sergio-correia , could you please point keylime sources to your PR repo&branch so we can see here how it goes? https://github.com/RedHat-SP-Security/keylime-tests/blob/main/plans/main.fmf#L8
> 
> Or maybe you did that in [keylime/keylime#1648](https://github.com/keylime/keylime/pull/1648) and tests were passing there and I have just missed that. If you feel confident to merge this PR we can proceed.

Hi, I added the temp commit here to verify it again, but yeah, I had changed it in the keylime PR before.

kkaarreell:
OK. So let's remove it again. :-) test results looks good and we can keep this PR opened or merge it right away, depending on what are your expectations regarding the merge of keylime upstream PR.

ansasaki:
https://github.com/keylime/keylime/pull/1648 was merged, so this should also be merged

PR #662, opened 2024-10-07
Do not add webhook server certificate to system CA trust store

kkaarreell:
Adding the CA certificate to 'trusted_server_ca' configuration should be sufficient.


kkaarreell:
No point waiting, current keylime is broken.

PR #670, opened 2024-10-31
Install builddeps and fix entering builddir

kkaarreell:


kkaarreell:
I think that IPv6 tests might be failing due to https://bugzilla.redhat.com/show_bug.cgi?id=2321325
Will try to implement a workaround.

PR #680, opened 2024-12-13
Add tests for `keylime-policy create measured-boot`

ansasaki:
Add sanity tests for `keylime-policy create measured-boot` command.

This also makes the tests that use the old `create_mb_refstate` script to use `keylime-policy` instead, when available.


functional/durable-attestion-sanity-on-localhost/test.sh
```diff
@@ -60,7 +60,11 @@ rlJournalStart
         rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
 
         # create refstat from fake binary_bios_measurements
-        rlRun "python3 /usr/share/keylime/scripts/create_mb_refstate /var/tmp/binary_bios_measurements mb_refstate.txt"
+        if which keylime-policy; then
```
sergio-correia, line 63: It's probably better to use `command -v` instead of `which` (also in the other places it is using it). Other than that, it looks great, thaniks.

ansasaki: I'll make the changes and push again

functional/durable-attestion-sanity-on-localhost/test.sh
```diff
@@ -60,7 +60,11 @@ rlJournalStart
         rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
 
         # create refstat from fake binary_bios_measurements
-        rlRun "python3 /usr/share/keylime/scripts/create_mb_refstate /var/tmp/binary_bios_measurements mb_refstate.txt"
+        if command -v keylime-policy; then
+            rlRun "keylime-policy -e /var/tmp/binary_bios_measurements -o mb_refstate.txt"
```
sergio-correia, line 64: It should be `keylime-policy create measured-boot -e -e /var/tmp/binary_bios_measurements -o mb_refstate.txt` here.

functional/keylime-policy-commands/test.sh
```diff
@@ -324,6 +330,30 @@ rlJournalStart
         popd ||:
     rlPhaseEnd
 
+    rlPhaseStartTest "Create measured boot policy"
+        mkdir -p measured-boot
+        pushd measured-boot || exit
```
kkaarreell, line 335: I don't expect this to fail but if it would, `exit` would mean that we won't do any clean up, which is not desired.

functional/keylime-policy-commands/test.sh
```diff
@@ -324,6 +330,30 @@ rlJournalStart
         popd ||:
     rlPhaseEnd
 
+    rlPhaseStartTest "Create measured boot policy"
+        mkdir -p measured-boot
```
kkaarreell, line 334: `rlRun "mkdir -p measured-boot"`

functional/measured-boot-policy-sanity/test.sh
```diff
@@ -56,27 +56,38 @@ rlJournalStart
         rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
     rlPhaseEnd
 
-    rlPhaseStartTest "Add agent with tpm_policy generated by create_mb_refstate script"
+    MB_POLICIES=()
+    rlPhaseStarttest "Create measured boot policy using different tools"
         # use installed create_mb_refstate from /usr/share/keylime/scripts
         rlRun "python3 /usr/share/keylime/scripts/create_mb_refstate $NO_SB_PARAM /sys/kernel/security/tpm0/binary_bios_measurements mb_refstate2.txt"
```
kkaarreell, line 62: Would it make sense to do this step also in an `if [ -f /usr/share/keylime/scripts/create_mb_refstate ]; then` block?

kkaarreell: Or maybe we want to be sure thatwe notice when the script is missing... ?

functional/keylime-policy-commands/test.sh
ansasaki: Right, I'll change it to make the test fail

functional/measured-boot-swtpm-sanity/main.fmf
```diff
@@ -6,7 +6,7 @@ description: |
  Uses accept-all measured boot policy for the verifier.
  Registers agent providing empty policy in --mb_refstate, confirms successful verification.
  Configure python agent to read custom bios_binary_measurements from /var/tmp.
- Re-registers agent providing --mb_refstate generated using create_mb_refstate.
+ Re-registers agent providing --mb_refstate generated using create_mb_refstate or keylime-policy.
```
kkaarreell, line 9: Hi @ansasaki , I think you have mixed two tests. We have `measured-boot-policy-sanity` which you are modifying above and which requires an actual measured boot (UEFI at least) and therefore is not run in keylime CI and there is also `measured-boot-swtpm-sanity` which uses faked `binary_bios_measurements` and is run in CI .. and whose `main.fmf` you are modifying here. I would say that changes you did (using both scripts) make sense primarily for the `swtpm` variant which gets run in CI. For the other test I think we can also use both scripts (since you already did the change).

functional/keylime-policy-commands/test.sh
```diff
@@ -216,8 +222,8 @@ rlJournalStart
     # Sign runtime policies.
 
     rlPhaseStartTest "Sign runtime policy with the ECDSA DSSE backend"
-        mkdir -p sign-ecdsa
-        pushd sign-ecdsa || exit
+        rlRun "mkdir -p sign-ecdsa"
+        pushd sign-ecdsa || rlFail "Failed to create sign-ecdsa directory"
```
kkaarreell, line 226: Won't hurt but given the `rlRun` around `mkdir` above the former command would fail already. `pushd` alone could fail in case the user is not able to enter a previously created dir, which I think could be due to permissions etc. Either way, you could use also `rlRun pushd` and avoid explicit `rlFail` call. Just FYI, no need to change it.

functional/measured-boot-swtpm-sanity/main.fmf
ansasaki: My bad! :sweat_smile: 
I'll make the changes to both tests.

functional/keylime-policy-commands/test.sh
ansasaki: I think it would be cleaner with all running using `rlRun` and no explicit failure. I'll make the changes.

functional/measured-boot-swtpm-sanity/main.fmf
ansasaki: Should be fixed now

functional/keylime-policy-commands/test.sh
ansasaki: Done

kkaarreell (approved):

PR #688, opened 2024-12-19
Update basic-attestation-on-localhost-api-version-bump for the agent with multiple API

ansasaki:
Replace the re-compilation of the agent using an older version with a simple
reconfiguration of  the agent enabling the desired API version.

### DIFF
--- a/compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
+++ b/compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
@@ -11,16 +11,7 @@ rlJournalStart
 
     rlPhaseStartSetup "Do the keylime setup"
         rlRun 'rlImport "./test-helpers"' || rlDie "cannot import keylime-tests/test-helpers library"
-        # install recommend devel packages from CRB if missing
-        rpm -q tpm2-tss-devel 2> /dev/null || INSTALL_PKGS="$INSTALL_PKGS tpm2-tss-devel"
-        rpm -q libarchive-devel 2> /dev/null || INSTALL_PKGS="$INSTALL_PKGS libarchive-devel"
-        if ! rpm -q zeromq-devel 2> /dev/null; then
-            rlIsRHEL '<10' && INSTALL_PKGS="$INSTALL_PKGS zeromq-devel"
-        fi
-        rlIsRHEL '<10' && EPEL_ARG="--enablerepo epel" || EPEL_ARG=""
-        [ -n "$INSTALL_PKGS" ] && rlRun "dnf --enablerepo \*CRB $EPEL_ARG -y install $INSTALL_PKGS"
         rlAssertRpm keylime
-
         # update /etc/keylime.conf
         limeBackupConfig
         # verifier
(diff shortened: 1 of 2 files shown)


compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
```diff
@@ -51,50 +42,22 @@ rlJournalStart
...
-        else
-            rlLogInfo "Will use agent sources from upstream repo"
-            rlRun "git clone ${RUST_KEYLIME_UPSTREAM_URL} ${WORKDIR}/rust-keylime"
-            rlRun "pushd ${WORKDIR}/rust-keylime"
+    rlPhaseStartTest "Get agent supported versions"
+        rlRun "limeStartAgent"
+        rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
+        mapfile -t SUPPORTED_VERSIONS< <(grep -ohE '> Starting server with API version.*' "$(limeAgentLogfile)" | grep -ohE '[0-9]*\.[0-9]*' | sort -V)
+        if [[ "${#SUPPORTED_VERSIONS[@]}" -lt 2 ]]; then
+            rlDie "Agent supports only one API version"
```
kkaarreell, line 50: Please, do not use `rlDie` as the test won't do a cleanup. Rather `rlFail` and conditionally skip other test phases.

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
```diff
@@ -51,50 +42,22 @@ rlJournalStart
...
-        BUILDDIR=$PWD
-        rlRun "cp ${BUILDDIR}/target/debug/keylime_agent /usr/bin/keylime_agent"
-        rlRun "popd"
     rlPhaseEnd
 
     rlPhaseStartTest "Add keylime agent with old API version"
+        rlRun "limeUpdateConf agent api_versions \"\\\"${OLD_VERSION}\\\"\""
         rlRun "limeStartAgent"
-        rlRun "limeWaitForAgentRegistration ${AGENT_ID}"
-        rlAssertGrep "Starting server with API version v${OLD_VERSION}" "$(limeAgentLogfile)" -E
```
kkaarreell, line 97: It would be nice to be able to verify that the agent is truly starting with the OLD version. Can we keep it in place or is the message no longer present?

kkaarreell (changes requested):
Thank you for test updates. I have two small change requests.

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
ansasaki: OK, I did not know that `rlDie` skipped the cleanup phase. I replaced it with `rlFail`

compatibility/basic-attestation-on-localhost-api-version-bump/test.sh
ansasaki: I re-added the check.

kkaarreell: I am not 100% sure but I think that the `limeWaitForAgentRegistration ` call was there to ensure that the grep is not executed too early. Let's see how tests will work.

kkaarreell (approved):
LGTM.

Koncpa (approved):

PR #689, opened 2024-12-20
Misc fixes

sergio-correia:
- functional/db-postgresql-sanity-on-localhost: fix psql setup
- Library/test-helpers: use -f when removing files


kkaarreell:
Hi @sergio-correia , I believe these changes are not necessary, the original issue has been addressed differently (I believe).
Can we close it?

sergio-correia:
Sure, we can close this, although I still think it sucks that the test is written in such a way that it does not work if we run it as root -- or has this been addressed in the meantime?

kkaarreell:
> Sure, we can close this, although I still think it sucks that the test is written in such a way that it does not work if we run it as root -- or has this been addressed in the meantime?

Tests can be run as root. The problem you were facing was that you have cloned a repo to /root. This is problematic in general as /root is using dedicated SELinux context and is more restricted compared to other (esp. temporary) directories. `tmt` uploads tests to /var/tmp so this is what tests are adjusted to and what is tested in CI.

PR #693, opened 2025-01-13
Enable compatibility tests

kkaarreell:


kkaarreell:
Unfortunately, I have no idea why the singlehost test is failing. Needs more investigation.

PR #696, opened 2025-01-15
Disable file.io, fix temp.sh

kkaarreell:


ansasaki (approved):

kkaarreell:
multihost test failure is due tu an unrelared selinux bug

PR #700, opened 2025-01-20
Set  the log level for the keylime lib as well

ansasaki:
Some of the log messages were moved from the keylime_agent to the keylime library.  With these changes, the messages logged by the keylime library are also visible.

This is a cherry pick of da8d7827ca65e6ea0630b3d77d809da5073b294f

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -2382,7 +2382,7 @@ limeconRunAgent() {
         PUBLISH_PORTS="-P"
     fi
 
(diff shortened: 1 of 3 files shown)


kkaarreell (approved):

ansasaki:
Apparently, the test error is due to a message output change, but the invalid signature is correctly detected and an exception is raised. This was previously fixed via 0e7afa6ab7cf1f4969f87ad50a5e76dcb75c6243

I'll merge this and prepare another PR for this cherry-pick

PR #721, opened 2025-02-17
Complete test migration to keylime-policy tool

kkaarreell:

PR #723, opened 2025-02-17
Complete test migration to keylime-policy tool

kkaarreell:
backport changes from 
https://github.com/RedHat-SP-Security/keylime-tests/pull/721


kkaarreell:
packit didn't catch up but the failure is no longer present  in https://github.com/RedHat-SP-Security/keylime-tests/pull/721 so I think it is safe to merge.

PR #740, opened 2025-02-19
flush TPM objects in tests

kkaarreell:


kkaarreell:
tested on an actual system with TPM device, PASS

PR #746, opened 2025-02-19
Cherry-pick changes for basic-attestation-on-localhost-api-version-bump from main

ansasaki:
This updates the test on `fedora-rawhide` to work with the agent with multiple API versions


kkaarreell:
The test Is not executed. Could you enable compatibility tests in distribution-fedora-keylime  test plan?

PR #758, opened 2025-02-19
Add option to double test duration

kkaarreell:
By defining `double_test_duration` context on `tmt` command line we can double test duration. This is handy especially for slow test systems.


sergio-correia (approved):

kkaarreell:
seems not to be needed anymore

PR #760, opened 2025-02-20
Enable sanity tests in distribution plans

kkaarreell:


kkaarreell:
test failure is not related.

PR #766, opened 2025-02-24
Introduce compatibility/api_version_compatibility

ansasaki:
This test runs the agent configuring each supported version separately and check that it passes attestation.


kkaarreell:
When the agent is stopped, the verifier will keep requesting quotes from it. I guess the new agent version starts soon enough so it won't failed the attestation but I am thinking which of the two options is better:

1. delete agent prior changing API version
2. do not delete the agent and let the verifier to detect version bump

compatibility/api_version_compatibility/test.sh
```diff
@@ -0,0 +1,88 @@
...
+        sleep 5
+        # start keylime_verifier
+        rlRun "limeStartVerifier"
+        rlRun "limeWaitForVerifier"
+        rlRun "limeStartRegistrar"
+        rlRun "limeWaitForRegistrar"
+        # create allowlist and excludelist
+        rlRun "limeCreateTestPolicy"
+
+        WORKDIR=$( mktemp -d -p "/var/tmp" )
```
kkaarreell, line 42: I think this is not needed as we are not running any scripts on the agent.
Also, line 85 is not needed.

ansasaki: Thank you! I'll remove these lines

ansasaki:
> When the agent is stopped, the verifier will keep requesting quotes from it. I guess the new agent version starts soon enough so it won't failed the attestation but I am thinking which of the two options is better:
> 
>     1. delete agent prior changing API version
> 
>     2. do not delete the agent and let the verifier to detect version bump

I think for now I would leave it with the version bump detection, but if it starts to cause failures, we can change it to delete the agent before adding again

kkaarreell:
Pls, remember to add this test also to a keylime upstream test plan. Thank you.

PR #767, opened 2025-02-27
Add setup/bootc_configure_kernel_ima_module

kkaarreell:


Koncpa (approved):
LGTM.

setup/bootc_configure_kernel_ima_module/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM localhost/bootc:latest
```
sergio-correia, line 1: What is this `localhost/bootc` image? Do we build it ourselves?

kkaarreell: This is the image used at that particular test system. With that we can build on top of it, without having to know a particular image.

PR #779, opened 2025-06-04
Enable logs generated by the keylime library

ansasaki:
The PR https://github.com/keylime/rust-keylime/pull/1018 moves code from the `keylime-agent` application to the `keylime` library.

This makes it necessary to set the log level for both the application and the library so that the messages are logged as expected.

This also changes the `sanity/keylime-secure_mount` test to ignore the source of the expected log.

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -479,8 +479,9 @@ __limeStartKeylimeService() {
 
     # if there is no unit file, execute the process directly
     else
-        # export RUST_LOG=keylime_agent=trace just in case we are using rust-keylime
-        RUST_LOG=keylime_agent=trace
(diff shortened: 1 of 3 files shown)


ansasaki:
> Hi @ansasaki , thank you for the update. Is the failing test due to a slower process start or was the expected message removed?

The expected messages were not removed, but since the code was moved in the PR (https://github.com/keylime/rust-keylime/pull/1018) to the `keylime` library, it is logged differently because the source of the log is different.

Basically, the `keylime_agent::` prefix in the source is replaced by the `keylime::` prefix (with pontential extra changes when the file that generated the log was renamed during the refactoring)

kkaarreell (approved):

PR #793, opened 2025-06-19
Adjust test scenario to use sign PQC algoritms

Koncpa:
Two current scenarios use as default sign algorithm RSA, adjust these to make it configurable to use RSA or ML-DSA-65. Part of this change is implementing function for key generation, current certgen library cannot generate PQ algorithm.


Koncpa:
It was manually tested for latest available system and working as expected.

Koncpa:
Scenarios were adjusted to use also ECDSA..

PR #796, opened 2025-06-26
Restore the use of x509KeyGen function

kkaarreell:


Koncpa:
LGTM! Rawhide is probably expected..

PR #798, opened 2025-07-04
keylime-secure-mount: Accept expected logs from the library

ansasaki:
After moving some code to the library, it is possible that some of the logs are generated there and not in the keylime_agent.

Therefore, drop the origin of the log from the expected log message.


sergio-correia (approved):

ansasaki:
The failing test is unrelated and is caused by some limit in the number of connections (QueuePool limit reached).

I'll ignore for now and merge this.

PR #800, opened 2025-07-11
Wait 3 seconds after starting the agent

kkaarreell:


kkaarreell:
c10s failure is due to https://github.com/keylime/keylime/issues/1689

PR #824, opened 2025-08-12
Fix pattern for keylime process detection

kkaarreell:
Processes like keylime_ima_emulator do not have / before its name in process listing.


kkaarreell:
failure not related, this is current openssl symbol issue.

PR #852, opened 2025-09-02
do not configure TPM with malformed EK on Image mode system

kkaarreell:


kkaarreell:
This needs more work. We cannot stop TPM with malformed certificate unconditionally in `limeStopTPMEmulator` because the service may not be configured.

PR #858, opened 2025-09-04
Disable malformed EK setup on ppc64le

kkaarreell:


kkaarreell:
test failure is expected

PR #882, opened 2025-09-15
Collect both RSA and ECC EK certificates with tpm2_getekcertificate

ansasaki:
In functional/add-agent-with-malformed-ek-cert test, collect both certificates separately when calling tpm2_getekcertificate.

In some systems, the tpm2_getekcertificate returns both certificates appended when called without specifying the output paths.


kkaarreell:
CI errors on Fedora are expected due to a selinux-policy issue.

PR #900, opened 2025-09-22
Use custom certificates for the agent too

ansasaki:
Previously, the agent supported only RSA key/certificate as the mTLS key because it was also used to encrypt data in the secure payload mechanism.

With https://github.com/keylime/rust-keylime/pull/1129, the agent is now able to use different algorithms for mTLS.

## Summary by Sourcery

Enable functional tests to generate and configure custom agent certificates for mTLS using the specified cryptographic algorithm

Tests:
- Generate agent key pair with CRYPTO_ALG instead of default RSA
- Sign the agent certificate with the intermediate CA
- Copy agent certificate and key into the test certificate directory
- Update agent configuration in tests to reference full paths for the server_cert and server_key

### DIFF
--- a/functional/basic-attestation-with-custom-certificates/main.fmf
+++ b/functional/basic-attestation-with-custom-certificates/main.fmf
@@ -39,6 +39,7 @@ extra-nitrate: TC#0611725
 /pqc_alg:
     environment:
         CRYPTO_ALG: mldsa65
+        AGENT_CRYPTO_ALG: ecdsa
     continue: false
     adjust+:
       - enabled: false
--- a/functional/basic-attestation-with-custom-certificates/test.sh
+++ b/functional/basic-attestation-with-custom-certificates/test.sh
@@ -8,6 +8,9 @@ WEBHOOK_SERVER_PORT=8980
 AGENT_ID="d432fbb3-d2f1-4a97-9ef7-75bd81c00000"
 MY_IP=127.0.0.1
 HOSTNAME=$( hostname )
+# When testing PQC, set the agent algorithm with something else due to
+# https://issues.redhat.com/browse/RHEL-117439
+[ -n "${AGENT_CRYPTO_ALG}" ] || AGENT_CRYPTO_ALG="${CRYPTO_ALG}"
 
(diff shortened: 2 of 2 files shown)


Koncpa (approved):

ansasaki:
> > ```
> >  [   FAIL   ] :: Command 'tail /var/tmp/limeLib/agent.log | grep 'Executing revocation action local_action_modify_payload'' (Expected 0, got 1)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This I can't explain as well, I can see the message in the agent's log. (Rawhide)
> > EDIT: I can, we are checking just last 10 lines, I will prepare a PR.
> > EDIT2: No, I can't, we are checking last 20 lines in fact :-(
> 
> So I did some cosmetic changes in #901 and I am not seeing the failure there. Still, it doesn't really explain why this is happening.

This one I do not understand, sorry.

kkaarreell:
I would also avoid changing payload mechanism. We need to keep it compatible with older releases so the current functionality would have to remain anyway.
Would you have an explanation why the issue is not visible on F41 and C10S?

ansasaki:
> I would also avoid changing payload mechanism. We need to keep it compatible with older releases so the current functionality would have to remain anyway. Would you have an explanation why the issue is not visible on F41 and C10S?

To be honest, the explanation I gave above still needs confirmation. I don't know why it is not visible there, but it is visible for fedora > 42.

kkaarreell:
> EDIT: The confirmation is not difficult. It would be enough to:
>     * Start verifier, registrar, agent; then add the agent to the verifier with the tenant
>     * Stop the agent while being attested and then start again (quick enough so that the verifier does not give up)
>     * If my theory is correct, the attestation will fail due to the new payload key hash extended into PCR#16

Event though the failure seems to be quite reliable in CI I wasn't equally successful in trying to reproduce it on my test system. I think that the timing is crucial here.

kkaarreell:
@ansasaki Hi, can we merge it? Test results looks good now. Thank you.

PR #981, opened 2025-12-17
Add regression/db-connection-leak-valgrind

sergio-correia:
This test uses valgrind and is based on the existing test
regression/db-connection-leak-reproducer.


regression/db-connection-leak-valgrind/main.fmf
```diff
@@ -0,0 +1,47 @@
...
+  - valgrind
+  - leak-detection
+require:
+  - yum
+  - sqlite
+  - coreutils
+  - valgrind
+recommend:
+  - keylime
+  - python3-tomli
```
kkaarreell: this is not needed as we are not running a revocation script

kkaarreell (changes requested):
I have just a few nitpicks. 
However, it seems that the test reports various connection leaks even for the current upstream code and not consistently.

PR #1070, opened 2026-04-23
Drop tests for features that are not implemented in RHEL-9.7.0

kkaarreell:
## Summary by Sourcery

Add helper utilities for configuring Keylime logging and add a new functional test covering tenant API version negotiation.

Enhancements:
- Extend configuration backup/restore helpers to include Keylime systemd drop-in directories and reload systemd when restored.
- Introduce helpers to enable and disable debug/trace logging for Keylime services, including Rust agents via systemd drop-ins.

Tests:
- Add a functional test that verifies tenant, verifier, registrar, and agent API version negotiation behavior under compatible and incompatible version configurations.

### DIFF
--- a/Library/test-helpers/lib.sh
+++ b/Library/test-helpers/lib.sh
@@ -372,7 +372,14 @@ Returns 0 when the initialization was successfull, non-zero otherwise.
 
 limeBackupConfig() {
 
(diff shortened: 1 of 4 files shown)


kkaarreell:
one test failure is currently expected

PR #1073, opened 2026-05-05
Add c10s prepare workaround for openssl version skew

sarroutbi:
The testing-farm-tag-repository can ship a stale openssl-devel that conflicts with a newer openssl-libs from the CentOS Stream 10 baseos repo, causing DNF dependency resolution to fail and aborting the entire prepare phase before any test runs.

Add an early prepare step (order 5) that syncs all openssl packages to the latest available version with --allowerasing, scoped only to RHEL-10 / CentOS Stream 10 via the existing adjust block. This runs before EPEL setup (order 10) and general package updates (order 30).

## Summary by Sourcery

Enhancements:
- Introduce an ordered prepare step to align OpenSSL-related packages and avoid DNF dependency resolution failures on RHEL 10 / CentOS Stream 10.

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -38,14 +38,15 @@ adjust:
           - dnf config-manager --set-disabled epel || true
   - when: distro == rhel-10 or distro == centos-stream-10
     prepare+:
-      # Workaround: testing-farm-tag-repository may ship a stale openssl-devel
-      # that conflicts with a newer openssl-libs from baseos. Sync all openssl
-      # packages to the latest available version before any other dnf operation.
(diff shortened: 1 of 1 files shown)


kkaarreell:
The issue has been addressed on TF side by setting a default priority for the tag repository.

sarroutbi:
> The issue has been addressed on TF side by setting a default priority for the tag repository.

Just let me keep the draft PR opened in case we need a quick workaround. Once addressed through TF, we can close it.

sarroutbi:
Not necessary any more

PR #1074, opened 2026-05-27
Disable test on RHEL-9.9 as well

kkaarreell:
## Summary by Sourcery

Tests:
- Update FMF metadata for the basic attestation on localhost API version bump compatibility test.


kkaarreell:
the failure is expected, the issue is not yet fixed in RHEL-9.9

PR #1088, opened 2026-07-07
Adjust rlWaitForCmd timeout and unify the definiton of tpm_check_peristent_handle()

4N0body5:
- Updated the fix of ambiguity on the s390x "tpm2_getcap ... | grep ..."   in the other script as well
-  Increased test timeout  in agent-service-start test:  HW TPM startup on aarch64 required 1s extra to pass as the scenario proceeded as expected

### DIFF
--- a/functional/push-authentication-cross-agent-attestation/custom-agent/test_tpm_helpers.sh
+++ b/functional/push-authentication-cross-agent-attestation/custom-agent/test_tpm_helpers.sh
@@ -24,9 +24,7 @@ SWTPM_PID=""
 tpm_check_persistent_handle() {
     local persistent_handle="${1}"
 
-    # On some architectures (s390x), grep -q exits early after finding a match,
-    # therefore we rather redirect the output to /dev/null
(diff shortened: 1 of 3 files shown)


functional/push-authentication-cross-agent-attestation/custom-agent/test_tpm_helpers.sh
```diff
@@ -24,9 +24,7 @@ SWTPM_PID=""
 tpm_check_persistent_handle() {
     local persistent_handle="${1}"
 
-    # On some architectures (s390x), grep -q exits early after finding a match,
-    # therefore we rather redirect the output to /dev/null
-    if tpm2_getcap handles-persistent 2>&1 | grep "${persistent_handle}" > /dev/null; then
```
kkaarreell, line 29: @4N0body5 you have modified the wrong function. "grep" is correct, "grep -q" is not.

kkaarreell (approved):
lgtm

PR #1099, opened 2026-07-22
keylime-secure-mount: Accept expected logs from the library

kkaarreell:
After moving some code to the library, it is possible that some of the logs are generated there and not in the keylime_agent.

Therefore, drop the origin of the log from the expected log message.

## Summary by Sourcery

Tests:
- Update keylime secure mount sanity tests to match log messages without asserting the keylime_agent module prefix.


kkaarreell:
test failure is unrelated

PR #1103, opened 2026-07-22
Fix SELinux module definition

kkaarreell:
On Rawhide loading of compiled module started to fail with:
  #semodule -i swtpm_permissive.pp'
  Failed to resolve typeattributeset statement at /var/lib/selinux/targeted/tmp/modules/400/swtpm_permissive/cil:2
  Failed to resolve AST
  semodule:  Failed!

## Summary by Sourcery

Bug Fixes:
- Resolve SELinux typeattributeset resolution failures when installing the swtpm_permissive compiled module on Rawhide.

### DIFF
--- a/setup/configure_tpm_emulator/swtpm_permissive.te
+++ b/setup/configure_tpm_emulator/swtpm_permissive.te
@@ -1,15 +1,18 @@
 policy_module(swtpm_permissive, 1.0)
-require {
+
 type swtpm_t;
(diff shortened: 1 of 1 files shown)


kkaarreell:
The selinux module seems to be fixed, however there is a problem on Rawhide with keylime/agent due to a new tpm2-tss. Keylime packages have to be rebuilt. 
bz#2504179 and bz#2504178

PR #1097, opened 2026-07-21
New test for postgresql pushmodel regression

4N0body5:
## Summary by Sourcery

Add a regression test ensuring Keylime push-model attestation succeeds when using a PostgreSQL backend and does not trigger agent ID truncation issues.

Tests:
- Add Beakerlib-based regression test that configures PostgreSQL and Keylime for push-model attestation and verifies successful agent attestation without StringDataRightTruncation errors.
- Introduce associated fmf metadata and PostgreSQL setup script to provision verifier and registrar databases for the regression scenario.

PR #1105, opened 2026-07-29
Update ALL kernels in IMA configuration setup

4N0body5:
On s390x, zipl may boot the running (+debug) kernel while grubby DEFAULT points elsewhere. Updating only DEFAULT leaves the booted entry without IMA cmdline args.
Using ALL ensures IMA args apply to whichever kernel actually reboots (not only DEFAULT).

See: https://artifacts.osci.redhat.com/testing-farm/2625a6de-4896-4b57-acf6-e1ef087372b4/

## Summary by Sourcery

Tests:
- Adjust kernel IMA configuration test to update all grubby kernels and verify arguments against the currently running kernel image.

### DIFF
--- a/setup/configure_kernel_ima_module/test.sh
+++ b/setup/configure_kernel_ima_module/test.sh
@@ -44,8 +44,10 @@ rlJournalStart
         rlRun "cat /proc/cmdline"
         rlRun "grubby --info ALL"
         rlRun "grubby --default-index"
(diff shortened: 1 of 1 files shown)


kkaarreell:
Hi. I don't think we should be updating all kernels leaving no option for recovery if something doesn't work. If there is a problem with debug kernel we should better address that particular issue. IIRC it is related to a compose bug, we could possibly include a workaround for that in the workarounds setup task.

kkaarreell:
obsoleted by https://github.com/RedHat-SP-Security/keylime-tests/pull/1097

PR #1106, opened 2026-07-29
Update ALL kernels in IMA configuration setup

4N0body5:
On s390x, zipl may boot the running (+debug) kernel while grubby DEFAULT points elsewhere. Updating only DEFAULT leaves the booted entry without IMA cmdline args.
Using ALL ensures IMA args apply to whichever kernel actually reboots (not only DEFAULT).

See failure example [here](https://reportportal-rhel.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#baseosqe/launches/all/2a1aef40-4e38-4845-9818-46bf7b5fb7c9): https://artifacts.osci.redhat.com/testing-farm/2625a6de-4896-4b57-acf6-e1ef087372b4/

## Summary by Sourcery

Ensure IMA kernel command-line arguments are applied to all bootable kernels rather than only the default, and verify they are present on the currently running kernel.

Enhancements:
- Update the IMA configuration test to use grubby --update-kernel ALL so IMA arguments are set for every kernel entry.
- Adjust test verification to check IMA arguments on the actually running kernel image instead of the grubby DEFAULT entry.

### DIFF
--- a/setup/configure_kernel_ima_module/test.sh
+++ b/setup/configure_kernel_ima_module/test.sh
@@ -44,8 +44,10 @@ rlJournalStart
         rlRun "cat /proc/cmdline"
         rlRun "grubby --info ALL"
         rlRun "grubby --default-index"
-        rlRun "grubby --update-kernel DEFAULT --args 'ima_appraise=${IMA_APPRAISE} ima_canonical_fmt ima_template=${IMA_TEMPLATE}'"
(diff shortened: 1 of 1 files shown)


kkaarreell:
obsoleted by https://github.com/RedHat-SP-Security/keylime-tests/pull/1097

PR #1107, opened 2026-07-31
Fix flaky --max-workers strace assertion on aarch64

4N0body5:
Count only process forks (SIGCHLD) from the main keylime-policy PID instead of all clone/clone3 lines, filtering out non-deterministic thread creation in child processes. Also widen the gap by comparing --max-workers 1 vs 5 (difference of 4 forks) instead of 1 vs 3 (difference of 2).

## Summary by Sourcery

Stabilize the keylime-policy --max-workers strace-based test by making its fork-counting logic deterministic and widening the expected difference in worker counts.

Bug Fixes:
- Prevent flaky behavior in the --max-workers functional test on aarch64 by ignoring non-deterministic thread clones when counting workers.

Tests:
- Update the strace-based functional test to count only SIGCHLD-producing forks from the main keylime-policy process and compare worker counts for 1 vs 5 workers instead of 1 vs 3.

Failure example [here](https://reportportal-rhel.apps.dno.ocp-hub.prod.psi.redhat.com/ui/#baseosqe/launches/all/46990/4852427).

### DIFF
--- a/functional/keylime-policy-commands/stall_server.py
+++ b/functional/keylime-policy-commands/stall_server.py
@@ -1,0 +1,38 @@
+"""HTTP server that serves repodata normally but stalls on RPM downloads.
+
+Used to test --max-workers by counting established TCP connections
+while RPM downloads are stalled.
+
+Usage: python3 stall_server.py [directory] [port]
+"""
+
+import http.server
+import os
+import sys
+import time
+
+
+class StallOnRpmHandler(http.server.SimpleHTTPRequestHandler):
+    def do_GET(self):
+        if self.path.startswith("/repodata/"):
+            super().do_GET()
+        else:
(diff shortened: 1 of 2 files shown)


kkaarreell:
This change turned out to be insufficient and I had to do additional test rewrite in https://github.com/RedHat-SP-Security/keylime-tests/pull/1102/changes

4N0body5:
> This change turned out to be insufficient and I had to do additional test rewrite in https://github.com/RedHat-SP-Security/keylime-tests/pull/1102/changes

Yes, it was my intention to backport the changes from that PR. My mistake, I did not notice it was merged as separate commits.

kkaarreell (approved):
Thank you.

PR #1118, opened 2026-08-06
Exclude push+PostgreSQL regression from agent upstream CI

4N0body5:
The test covers verifier/registrar PostgreSQL schema behavior, not agent functionality, so it should not be required in rust-keylime CI parity.

## Summary by Sourcery

CI:
- Update upstream agent CI configuration to skip the push-model attestation PostgreSQL regression test that is not specific to the agent.


kkaarreell (approved):
LGTM.

PR #1119, opened 2026-08-06
Exclude push+PostgreSQL regression from agent upstream CI

4N0body5:
The test covers verifier/registrar PostgreSQL schema behavior, not agent functionality, so it should not be required in rust-keylime CI parity.

## Summary by Sourcery

Tests:
- Adjust functional test configuration to stop running the PostgreSQL push-model attestation regression test in the agent upstream CI.


kkaarreell:
This PR is not necessary, upstream tests are run only from the main branch

